PR #28, opened 2020-10-04
Make libfuse optional

### LINKED ISSUE #26: Remove dependency on native libfuse library
This can be done by reimplementing the mount and unmount functions from libfuse, including the IPC with fusemount binary

cberner:
Fixes #26

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Install packages
         run: |
           sudo apt update
-          sudo apt install -y libfuse-dev build-essential
+          sudo apt install -y libfuse-dev libfuse3-dev build-essential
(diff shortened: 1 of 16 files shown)


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - name: Install packages
         run: |
           sudo apt update
-          sudo apt install -y libfuse-dev build-essential
+          sudo apt install -y libfuse-dev libfuse3-dev build-essential
```
Mic92, line 15: Do we need this in our ci container as well?

cberner: If you build with libfuse (which is turned on by default), then yes. You have to build without default features to disable it

PR #33, opened 2020-10-12
Add feature-flagged serde support

pato:
Introduces a new `serializable` feature flag that when enabled will auto-derive serde support for `FileType` and `FileAttr`. Also bumps the version to `0.4.1`.

The motivation for this is to be facilitate use-cases that require serializing (either for persistence or network transport) the aforementioned structs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 * Add ABI 7.21
 * Add ABI 7.22
 
+## 0.4.1 - 2020-10-07
+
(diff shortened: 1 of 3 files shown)


pato:
@cberner - I have tentatively bumped the version to `0.4.1`, but this is currently targeting master (which I believe is where you are preparing the `0.5` work). Perhaps you can create a branch off the published `0.4` version and we can re-target this PR to that? We might also want to change the date in the changelog for the release once we have one. Let me know if there is anything that you'd like changed

cberner:
Cool, looks good! I'll merge this to master, but then backport it and release off the 0.4 commit

PR #37, opened 2020-10-13
Add tests for mounting without linking libfuse on OSX

cberner:


cberner:
loading the osxfuse kernel extension isn't supported: https://github.com/travis-ci/travis-ci/issues/10017

PR #63, opened 2020-10-24
Documents how to use Nix packages.

rodrigovalin:
I've playing around with fuser and I'm willing to give it a try to implement a new tool. This is my first contribution to use Nix packages instead of Homebrew.

Please let me know if there's anything here I should add or improve.

> This is the *kernel* part that needs to be installed always.

I added this small comment because I actually missed this part!

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,9 +62,9 @@ sudo yum install fuse-devel pkgconfig
 
 ### macOS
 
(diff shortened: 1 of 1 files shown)


cberner:
Merged. Thanks!

Issue #102, opened 2020-11-27
Use Slightly Higher Level Types


grahamc:
A lot of places in the code use base level types to mean lots of different things. For example, ReplyCreated:

```
pub fn created(
    self,
    ttl: &Duration,
    attr: &FileAttr,
    generation: u64,
    fh: u64,
    flags: u32
)
```

I propose creating types called GenerationId and FileHandleID etc. which are `.into()`d when passed to the kernel. This way, it is more difficult to confuse arguments and values by mistake. This is especially true when calling the created function, since you don't even have the benefit of naming the argument being passed. What I mean by this is constructing a struct it would be easier to tell the mistake:

```
Foo {
  generation: fh,
  fh: generation
}
```

vs:

```
created(duration, attr, fh, generation, flags)
```

cberner:
Thanks for the suggestion. I think this is a good idea. Although, if I understand correctly it would be a breaking change to the `Filesystem` API. I don't want to introduce breaking changes unless they're necessary. I'll consider changing this if I make a bigger update to that API though. I'd like to add support for `async` at some point, and I'm sure that will result in a large change

siscia:
Since we are talking about changes to the API, I was working on an API that returns a `Result<Something, ErrorCode>` instead of using the (I think) awkward `Reply` API.

wmanley:
I created such newtypes in #119, but they're not yet exposed as public API. I'm working towards that in #136, tracking issue: #138.

cberner:
Closing due to inactivity, but feel free to re-open if you're still working on this.

PR #103, opened 2020-11-29
sketch for new API

siscia:
Of course incomplete, but this is what I was referring to in #102 

I am now wondering the `Inode.into()` method runtime cost. I would guess zero, but it would be good to have confirmation.

Similarly, for the Generations, File, Directory handles.

If there is interest, I can finish implementing the API.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -896,13 +896,13 @@ where
     fn getattr(&mut self, req: &Request<'_>, ino: u64, reply: ReplyAttr) {
         match self.getattr(req, ino.into()) {
             Ok((ttl, attr)) => reply.attr(ttl, attr),
-            Err(n) => reply.error(n.into()),
+            Err(n) => reply.error(n),
         }
     }
     fn readlink(&mut self, req: &Request<'_>, ino: u64, reply: ReplyData) {
         match self.readlink(req, ino.into()) {
             Ok(d) => reply.data(d),
-            Err(n) => reply.error(n.into()),
+            Err(n) => reply.error(n),
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -762,6 +762,61 @@ pub trait Filesystem {
     }
 }
 
+#[derive(Debug, Hash, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct Inode(u64);
+
+impl From<u64> for Inode {
+    fn from(ino: u64) -> Self {
+        Self(ino)
+    }
+}
+
+pub trait Filesystem2 {
+    fn init(&mut self, _req: &Request<'_>, _config: &mut KernelConfig) -> Result<(), c_int> {
+        Ok(())
+    }
+    fn destroy(&mut self, _req: &Request<'_>) {}
+    fn forget(&mut self, _req: &Request<'_>, _ino: Inode, _nlookup: u64) {}
+    fn getattr(
+        &mut self,
+        _req: &Request<'_>,
+        _ino: Inode,
+    ) -> Result<(&std::time::Duration, &FileAttr), c_int> {
+        Err(ENOSYS)
+    }
+    fn readlink(&mut self, _req: &Request<'_>, _ino: Inode) -> Result<&[u8], c_int> {
+        Err(ENOSYS)
+    }
+}
+
+impl<T> Filesystem for T
+where
+    T: Filesystem2,
+{
+    fn init(&mut self, req: &Request<'_>, config: &mut KernelConfig) -> Result<(), c_int> {
+        self.init(req, config)
+    }
+
+    fn destroy(&mut self, req: &Request<'_>) {
+        self.destroy(req)
+    }
+    fn forget(&mut self, req: &Request<'_>, ino: u64, nlookup: u64) {
+        self.forget(req, ino.into(), nlookup)
+    }
+    fn getattr(&mut self, req: &Request<'_>, ino: u64, reply: ReplyAttr) {
+        match self.getattr(req, ino.into()) {
+            Ok((ttl, attr)) => reply.attr(ttl, attr),
+            Err(n) => reply.error(n),
+        }
+    }
+    fn readlink(&mut self, req: &Request<'_>, ino: u64, reply: ReplyData) {
+        match self.readlink(req, ino.into()) {
+            Ok(d) => reply.data(d),
+            Err(n) => reply.error(n),
+        }
+    }
+}
+
 /// Mount the given filesystem to the given mountpoint. This function will
 /// not return until the filesystem is unmounted.
 ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -771,8 +771,95 @@ impl From<u64> for Inode {
     }
 }
 
+// we are missing most of the error codes provided by linux
+// if necessary, add the necessary.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -762,6 +762,61 @@ pub trait Filesystem {
...
+    fn init(&mut self, _req: &Request<'_>, _config: &mut KernelConfig) -> Result<(), c_int> {
+        Ok(())
+    }
+    fn destroy(&mut self, _req: &Request<'_>) {}
+    fn forget(&mut self, _req: &Request<'_>, _ino: Inode, _nlookup: u64) {}
+    fn getattr(
+        &mut self,
+        _req: &Request<'_>,
+        _ino: Inode,
+    ) -> Result<(&std::time::Duration, &FileAttr), c_int> {
```
cberner, line 784: These references seem like they will make the implementation complicated. Have you tried this interface out to see out well it works?

cberner (commented):
I generally like this approach of switching to functions that return `Result`, but returning references seems like it will make the implementations complicated. What do you think?

Also, instead of returning `c_int` as the error type, I think it would be better to return an enum which can then be mapped to the correct libc error code. There are some cases where that mapping is platform dependent, so handling it in the library seems like it will make life easier for implementors

src/lib.rs
siscia: Yep, you are right! I tried before a similar API, but not with the reference.

In this case makes much more sense to get the value.

siscia:
For the errors, I am not sure, I am afraid it will explode in the number of variants.
I was hoping to use `std::io::Error` but it is clearly not compressive: https://doc.rust-lang.org/std/io/enum.ErrorKind.html

Do you have any example of some other library that wrapped the possibles errors? I checked `libc` but there is nothing.

The advantage of using `c_int` is extreme simplicity.

cberner:
Nothing complete. I created something similar in the [filesystem](https://github.com/fleetfs/fleetfs/blob/84fe3476681c268afcb69d039b1cb3cf4dbd56c7/src/fuse_adapter.rs#L119) that I'm working on. It's definitely not complete though.

The main reason I thought about it is that the error code for functions like `getxattr` are platform dependent, and it would be nice to abstract that: https://github.com/cberner/fuser/blob/c32aa8fc625c905565b8706ad91a68f496658bf1/examples/simple.rs#L1452

src/lib.rs
cberner: Ya, I think in most cases it's best to return a value. Although, I think `read` and maybe `readlink` would need to return references as a potential optimization to avoid copying memory. Can you think of a better way to deal with that?

I rely on that in fleetfs, so that the TCP client can reuse a single thread local buffer: https://github.com/fleetfs/fleetfs/blob/84fe3476681c268afcb69d039b1cb3cf4dbd56c7/src/fuse_adapter.rs#L661

siscia:
Hum, I am not sure about making ourself an error type being a good idea.
I believe there are way too many possible errors, and we will never be compressive, and eventually all those errors need to be converted into `c_int`.

Which give me the idea, that maybe we can get the output as `Into<c_int>`

Not sure if it is possible or convenient...

cberner:
[This list](https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/linux/mod.rs#L2534) in `libc` looks like it covers most of the error codes, and it doesn't look that long to me. There are some beyond that list, so I agree the complete list is long, but most of them will never be used. 
I think the best thing to do is to create an error enum with a `custom` value which can be set to an arbitrary `c_int`, but then have defined enum values for the common error codes. For example, I took this approach for mount options: https://github.com/cberner/fuser/blob/5b77bf1a041cbc8d2e1001718b9dd621cd6f5c39/src/mount_options.rs#L16

Thoughts?

siscia:
That looks not that bad at all!

Let me prepare another sketch!

src/lib.rs
```diff
@@ -762,6 +762,151 @@ pub trait Filesystem {
...
+impl From<u64> for Inode {
+    fn from(ino: u64) -> Self {
+        Self(ino)
+    }
+}
+
+// we are missing most of the error codes provided by linux
+// if necessary, add the necessary.
+// https://man7.org/linux/man-pages/man3/errno.3.html
+pub enum Errno {
```
cberner, line 777: This looks good, but I think we should use more descriptive error names, so instead of `EFBIG` something like `FileTooLarge`

siscia:
Sorry, I don't believe we are on the same page about the end result.

Of course feel free to continue on somehow similar lines.

wmanley:
Regarding errno: I've found that the newtype pattern works quite well in this instance.  I have:

```rust
#[derive(Copy, Clone, Debug)]
#[repr(transparent)]
struct Errno(libc::c_int);
impl Errno {
    const EFBIG: Errno = Errno(libc::EFBIG);
    const EIO: Errno = Errno(libc::EIO);
    const ENOENT: Errno = Errno(libc::ENOENT);
    const ENOTDIR: Errno = Errno(libc::ENOTDIR);
    const EISDIR: Errno = Errno(libc::EISDIR);
    const ENODATA: Errno = Errno(libc::ENODATA);
}
impl From<std::io::Error> for Errno {
    fn from(x: std::io::Error) -> Self {
        match x.raw_os_error() {
            Some(errno) => Errno(errno),
            None => Errno::EIO,
        }
    }
}
```

It has the same representation as the real errno, so you can losslessly convert to and from it, but unlike c_int it makes sense to implement `From<YourError> for Errno`. This way you can have a function that returns `Result<(), Errno>` and the `?` machinery will do the appropriate conversions.

I've also been using a macro to help with this error handling:

```rust
/* This is like the old try! macro, but for the reply pattern that fuse-rs
 * employs. */
macro_rules! try_reply {
    ($reply:expr, $stmt:expr) => {
        match $stmt {
            Ok(x) => x,
            Err(err) => {
                let err: Errno = err.into();
                $reply.error(err.0);
                return;
            }
        }
    };
}
```

So I can write code like:

```rust
        let mut f = try_reply!(reply, File::open("foo.txt"));
        let mut v = vec![];
        let attr = try_reply!(reply, f.read_to_end(&mut v));
        reply.data(&*v);
```

cberner:
Ah ya, I think that's a good approach. Do you have any thoughts on error codes that are platform dependent like for the xattr methods? I think it'd be nice to have a single error code for those, and then translate it into the correct errno: https://github.com/cberner/fuser/pull/103#issuecomment-736230841

wmanley:
Constants might work well here. Something like:

```rust
#[cfg_if(target_os=="mac")]
const NO_XATTR : Errno = Errno::ENOATTR;

#[cfg_if(target_os=="linux")]
const NO_XATTR : Errno = Errno::ENODATA;
```

Or maybe have a specific error enum for this that is convertible to `Errno`, but with platform specific behaviour:

```rust
enum XattrError {
    XAttrNotFound,
    BlahBlah
}

impl From<XAttrError> for Errno {
...
}
```

This would be useful for creating the errors, but not so useful for matching on them.

It's not something I've put much thought into as I'm only targetting linux in my use of this crate.

cberner:
Oh ya, that seems like it'll work fine. This approach looks good to me! Are you interested in sending a PR? If not, I'll probably get around to it in a few weeks

wmanley:
Yes, I'll send a PR.  I should have some time for it next week.

PR #104, opened 2021-01-07
autodetect installed macfuse version 4 in build.rs

mike-kfed:
I was redirected here in my issue report for zargony/fuse-rs#155 when building the crate against macFUSE 4.x. There were lots of renames going on, one of them being the pkg-config information. This PR autodetects an installed v4 and makes it build against v4 - the hello.rs example plus tests work fine for me (tested on macOS 11.1 with macFUSE 4.0.5 on a M1 MacBook Air).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,5 +1,3 @@
-static MAC_FUSE_4: &str = "/Library/Filesystems/macfuse.fs/Contents/Resources/mount_macfuse";
-
 fn main() {
     #[cfg(all(not(feature = "libfuse"), not(target_os = "linux")))]
     unimplemented!("Building without libfuse is only supported on Linux");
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -6,9 +8,14 @@ fn main() {
     {
         #[cfg(target_os = "macos")]
         {
+            let probelib = if std::path::Path::new(MAC_FUSE_4).exists() {
```
cberner, line 11: Checking for the existence of this path seems brittle to me. Is it documented anywhere that it's guaranteed to exist for MacFuse 4.0? I think it would be better to detect this via pkg_config. In other words, first try to link against "fuse" and then fallback to "osxfuse", just like the non-macos code path below: https://github.com/cberner/fuser/pull/104/files#diff-d0d98998092552a1d3259338c2c71e118a5b8343dd4703c0c7f552ada7f9cb42L23

cberner (changes requested):
Thanks for the contribution! I had one comment that I think will make the detection of macfuse more robust

build.rs
mike-kfed: my reasoning was that some package management tools might do some symlink shenanigans with the pkg_config files,  so I decided to test for a file that only v4 installs. But you are totally right, simply trying to use it is a more stable approach - I implemented your suggested changes.

cberner (approved):
Looks good to me. Thanks for the contribution!

PR #106, opened 2021-01-09
add_capabilities method for KernelConfig.

yzyDavid:
good idea, I'll modify that as you wish.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -134,8 +134,8 @@ pub struct KernelConfig {
 }
 
 impl KernelConfig {
-    fn new(capabilities: u32, max_readahead: u32) -> KernelConfig {
-        KernelConfig {
+    fn new(capabilities: u32, max_readahead: u32) -> Self {
+        Self {
             capabilities,
             requested: INIT_FLAGS,
             max_readahead,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -153,6 +153,19 @@ impl KernelConfig {
...
     }
 
+    /// Add a set of capabilities.
+    ///
+    /// On success returns Ok, else return 1 when capabilities you provided are not all supported by kernel.
+    pub fn add_capabilities(
+        self: &mut Self,
+        capabilities_to_add: u32) -> Result<(), u16> {
+        if capabilities_to_add & self.capabilities != capabilities_to_add {
+            return Err(1);
```
cberner, line 163: Instead of returning `1` on error, can you return the set of requested capabilities that can't be supported? i.e. `capabilities_to_add - (capabilities_to_add & self.capabilities)`. I think that will be more useful to people who need to handle the error

cberner (changes requested):
Thanks for the contribution! I had one comment, but otherwise this looks good

yzyDavid:
I'll close this PR and open a new one.

PR #113, opened 2021-02-19
Soft-deprecate FileAttr::padding

wmanley:
This was added in 7f677ce34642870b36b2bb73c952762b42c26eac, but I think it
was a mistake.  AFAICS the padding is required by the FUSE ABI, but isn't
actually used, so it shouldn appear in `fuse_abi::fuse_attr`, but not in
in our `FileAttr` API.

Removing it now would be a breaking change, so I'll just ignore it and
document that it's unused.  It can be removed for 0.8.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -130,7 +130,7 @@ pub struct FileAttr {
     pub rdev: u32,
     /// Block size
     pub blksize: u32,
-    /// Padding
+    /// Unused - this will be removed in the future.
     pub padding: u32,
     /// Flags (macOS only, see chflags(2))
     pub flags: u32,
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -115,7 +115,7 @@ fn fuse_attr_from_attr(attr: &FileAttr) -> fuse_attr {
         #[cfg(feature = "abi-7-9")]
         blksize: attr.blksize,
         #[cfg(feature = "abi-7-9")]
-        padding: attr.padding,
+        padding: 0,
     }
 }
 
@@ -144,7 +144,7 @@ fn fuse_attr_from_attr(attr: &FileAttr) -> fuse_attr {
         #[cfg(feature = "abi-7-9")]
         blksize: attr.blksize,
         #[cfg(feature = "abi-7-9")]
-        padding: attr.padding,
+        padding: 0,
     }
 }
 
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -130,7 +130,7 @@ pub struct FileAttr {
     pub rdev: u32,
     /// Block size
     pub blksize: u32,
-    /// Padding
+    /// Unused - this will be removed in the future.
```
cberner, line 133: Good catch, I mistook this for a real field when I added 7.9 support :(

Can you add `#[deprecated]` so that people get a compiler warning? https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html

Other wise this looks good to me.

cberner (changes requested):

src/lib.rs
wmanley: I didn't do that as you still need to fill in padding when constructing the struct, and there's no way to construct a `FileAttr` without including `padding: 0`. This causes rustc to complain: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=pub%20struct%20Moo%20%7B%0A%20%20%20%20hi%3Au64%2C%0A%20%20%20%20%23%5Bdeprecated%5D%0A%20%20%20%20padding%3A%20u32%2C%0A%7D%0A%0Afn%20main()%20%7B%0A%20%20%20%20Moo%20%7Bhi%3A%203%2C%20padding%3A%200%7D%3B%0A%7D). It seems a little unfair to force deprecation warnings on our users with no way of fixing them in a point release.

I do have some more changes brewing that help constructing `FileAttr` instances though.

cberner: I see, ya good point. In that case, how about we just remove the field now? 0.7 is already released, so I think this field can be removed and that'll go into the 0.8 release

wmanley: IMO backwards compatibility is very important, and any API breaking changes need to have sufficient benefit to justify asking downstream projects to upgrade. In this case IMO the benefit isn't worth 0.8 yet.  I was thinking 0.7.1 or similar would be a good next release, and this can be removed for 0.8 a bit later.

wmanley: I've also got some other changes that would require API breaks that I'd like to get into a 0.8 breaking API release, but they're not ready yet.

cberner:
Ok. I'll merge this for now. The next release is probably going to be 0.8 though, since there's a pr to switch Filesystem to be async #112

cberner (approved):

cberner:
Closing in favor of #145

PR #114, opened 2021-02-19
Remove unsafe from reply.rs

wmanley:
Instead we use the zerocopy crate, which should be safer.  The code here
is a lot more straightforward, particularly around `dirent` handling.

Note the previous implementation was unsound: Rust doesn't guarantee a
memory layout for tuples.  Also the previous implementation of `as_bytes`
was unsound.  It could be applied to structs with padding or uninitialised
memory.

AFAIU this is not a breaking change because `ReplyRaw` is not exported by
`lib.rs` so already cannot be used externally.


cberner:
Merged. Thanks!

PR #115, opened 2021-02-21
Remove unsafe from the request parsing code

wmanley:
Instead we use `zerocopy::FromBytes`.

Previously the code did not ensure alignment, which could lead to undefined
behaviour.  Now we do check alignment I've had to fix it in a few places
using `AlignedData<[u8; sz]>` as a buffer in tests.  Unfortunately we can't
use `Box<AlignedData<u8; BUFFER_SIZE>>` as our main buffer because it's too
large to create on the stack and Rust currently offers no way to allocate
it directly on the heap.  Instead we have to fiddle with offsets for the
`Vec`.

### DIFF
--- a/src/ll/argument.rs
+++ b/src/ll/argument.rs
@@ -4,7 +4,6 @@
 //! structures (request arguments).
 
 use std::ffi::OsStr;
-use std::mem;
 use std::os::unix::ffi::OsStrExt;
 
 /// An iterator that can be used to fetch typed arguments from a byte slice.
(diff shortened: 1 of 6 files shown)


wmanley:
@cberner: I consider this mergable. Please let me know if you want any changes.  There are no API changes here.

cberner:
Great! I'll take a look soon

src/ll/argument.rs
```diff
@@ -30,42 +29,50 @@ impl<'a> ArgumentIterator<'a> {
...
 
-    /// Fetch a slice of bytes of the given size. Returns `None` if there's not enough data left.
-    pub fn fetch_bytes(&mut self, amt: usize) -> Option<&'a [u8]> {
-        if amt > self.data.len() {
-            return None;
+    /// Fetch a typed argument. Returns `None` if there's not enough data left.
+    pub fn fetch<T: zerocopy::FromBytes>(&mut self) -> Option<&'a T> {
+        match zerocopy::LayoutVerified::<_, T>::new_from_prefix(self.data) {
+            None => {
+                if self.data.as_ptr() as usize & (core::mem::align_of::<T>() - 1) != 0 {
```
cberner, line 36: I think it would be more clear to check `self.data.as_ptr() as usize % core::mem::align_of::<T>() != 0`. Technically, `align_of` doesn't guarantee that it returns a power of 2 -- although I assume it does on all platforms.

src/session.rs
```diff
@@ -91,12 +91,13 @@ impl<FS: Filesystem> Session<FS> {
     pub fn run(&mut self) -> io::Result<()> {
         // Buffer for receiving requests from the kernel. Only one is allocated and
         // it is reused immediately after dispatching to conserve memory and allocations.
-        let mut buffer: Vec<u8> = Vec::with_capacity(BUFFER_SIZE);
+        let mut buffer = vec![0; BUFFER_SIZE];
+        let buf = aligned_sub_buf(buffer.deref_mut(), 8);
```
cberner, line 95: Can you replace the `8` with `align_of::<fuse_in_header>`? I think that will be more clear, and constants like this make me nervous that they will break one day when moving to a new platform -- e.g. u128 is 16 byte aligned.

cberner (approved):
Had a few minor comments. Otherwise this looks good though! Let me know when they're fixed and I'll merge this

src/session.rs
cberner: oh, actually I'm wrong. u128 alignment is 8, not 16. On a theoretical platform in the future, it seems like it might be though

src/ll/mod.rs
```diff
@@ -3,4 +3,23 @@
...
 
 mod request;
+use std::ops::{Deref, DerefMut};
+
 pub use request::{Operation, Request, RequestError};
+
+/// If we want to be able to cast bytes to our fuse C struct types we need it
+/// to be aligned.  This struct helps getting &[u8]s which are 8 byte aligned.
+#[repr(align(8))]
+pub struct AlignedData<T>(pub T);
```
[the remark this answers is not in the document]
wmanley: I've moved it into:

```
#[cfg(test)]
mod tests {
```

so I didn't need to add `#[cfg(test)]` to the `use` and the `impl` as well.

src/ll/argument.rs
wmanley: Done. I think technically speaking it [must be a power of 2](https://doc.rust-lang.org/reference/type-layout.html#size-and-alignment)

> Alignment is measured in bytes, and must be at least 1, and always a power of 2.

but I agree that using `%` is much clearer and will be optimised down to the same thing anyway.

src/session.rs
wmanley: Done.

Technically it's not `align_of::<fuse_in_header>` we want here. We want the max of the alignments of all the structures we'll be casting to, but in practice this will never be a problem, and what you suggest reads a lot better.

wmanley:
Thanks for the review.  I've fixed the issues and force-pushed.

cberner:
Merged. Thanks!

PR #116, opened 2021-02-24
Adding a macos github actions ci step

ianoc:
This is a reproduction of the appveyor ci checks for macos. The reason is the appveyor is slow, but also they kill/fail tests rather than any sort of queue if push more than a few commits.

Making this purely additive as a new check to start with, then could remove appveyor if unneeded.

### DIFF
--- a/.github/workflows/macos_ci.yml
+++ b/.github/workflows/macos_ci.yml
@@ -1,0 +1,49 @@
+name: Macos CI
+
+on:
+  push:
+
+jobs:
+  ci-macos:
+    runs-on: macos-latest
+    env:
+      PKG_CONFIG_PATH: /usr/local/lib/pkgconfig:$PKG_CONFIG_PATH
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install packages
+        run: |
+          brew update > /dev/null && brew install --cask osxfuse
+      - name: Start fuse
+        run: |
+          /Library/Filesystems/osxfuse.fs/Contents/Resources/load_osxfuse
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          components: rustfmt, clippy
+      - name: Cache cargo registry
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/registry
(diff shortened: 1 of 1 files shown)


.github/workflows/macos_ci.yml
```diff
@@ -0,0 +1,49 @@
+name: Macos CI
+
+on:
+  push:
```
cberner, line 4: can you change to `pull_request`? It looks like these tests didn't run, and I think the `osx_mount_tests.sh` test will fail. I tried to run it on Github Actions before, but they didn't support installing kernel extensions, which is why I used Appveyor

cberner (changes requested):

.github/workflows/macos_ci.yml
ianoc: i don't think anything will work per say on your repo unless you enable some flag --to avoid malicious stuff if i update the action they don't run here afaik

ianoc: https://github.com/ianoc/fuser/runs/1974135044?check_suite_focus=true

should be the same run here.

(This change means tests will run after merges to master and if you push a branch before opening a PR. it should strictly afaik run tests more often than on pull request only)

ianoc: -- this is mostly handy so with a fork you get CI with your own github actions that i've seen before you open a PR against the original repo

cberner: > i don't think anything will work per say on your repo unless you enable some flag --to avoid malicious stuff if i update the action they don't run here afaik

Ahhh ok, I didn't realize that's why it didn't run. That makes sense though

cberner (approved):

cberner:
Merged. Thanks!

.github/workflows/macos_ci.yml
ianoc: Yep, this stuff is a little non-obvious. When i edited it the actions in the other PR thats why it all shut down on it. I believe there was some security exploits about trying to exfiltrate permissions of the host repo at some point is probably why

PR #117, opened 2021-02-25
Split up integration testing, and try cache some state

ianoc:
This one is to try take the really long test which seems to have random kills or out of disk space issues for me and make it both more reliable and also much much faster. Takes 36mins now for the slowest segment. (Driven heavily by one test we could disable too if we felt like it)

### DIFF
--- a/.github/scripts/xfs_inside_docker_tests.sh
+++ b/.github/scripts/xfs_inside_docker_tests.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+set -e
+
+# Check for rustup installed
+if [ ! -f ~/.rustup/settings.toml ]; then
+ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain=1.47.0
+fi
+export PATH=/root/.cargo/bin:$PATH
+
+if [ ! -f /code/fuse-xfstests/check ]; then 
+    mkdir -p /code
+    cd /code 
+    git clone https://github.com/fleetfs/fuse-xfstests
+    cd fuse-xfstests
+    git checkout 0166199783962f0d988dfc5fbfea6aba4ac9143f
+    make
+fi
+
+cd /code/fuser
+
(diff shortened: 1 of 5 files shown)


ianoc:
Rebased on master, failing test was just the appveyor being out of minutes/too busy i believe.

ianoc:
To view the actions as they are running here you'll need to visit the commit in my fork

https://github.com/ianoc/fuser/commit/3877e6e1127b010d6a74f6fe05e4698695b4b2d5 (should be the same sha as listed here).

Then the ci icon should be by the commit title, which will list all the jobs.

There will be a good few more.


Things to look for:

I split the make tests into 3 types of things:

- The pre step, since building/installing cargo deny can take a few minutes
- The non-xfstests other tests from the make file
- The xfs tests contain subsections of the generic based on wild card/names.
   - Here its good to check against https://github.com/fleetfs/fuse-xfstests/tree/gocryptfs-2018-08-18/tests/generic to ensure the ranges being ran are inclusive. I included effectively 199 extra spots based on what i can see in there.
   - Click into one or 2 of these sub and make sure it looks like its running that sub section, it seems to spew a little nonsense that the `generic/001.bad` files aren't tests, which seems just harmless noise. Then from what i can see it does go on to run all the tests in the section as expected.


Upside of this change:

-> I think it should be more stable since disk pressure or any cross talk inside the Makefile tests will be relieved
-> It completes far faster, ~36mins for all tests, and since the tests run in parallel we should get feedback about test failures much more rapidly. 


Downside:
-> Forking of some testing behavior (limited to xfstests where we have to be a bit more invasive with our caching/handling)
-> More code/complexity relatively


Since make test still works as it did before, this should be a change we can revert back out of relatively easily in the future if it seems like a bad idea/more maintenance if keeping the Makefile and the list of sub tests in github actions in sync is a lot of work.

cberner (approved):
Having, now, 4 different Dockerfiles for the tests is unfortunate. Faster tests is great though, and we can always clean that up later

PR #118, opened 2021-02-26
The github actions depend on their being a lock file for stability

ianoc:
The github actions (old and new ones) all rely on the hash of a `Cargo.lock` file, the ci broke on master i believe because there is no lock file and its causing some cache pollution


ianoc:
Well my understanding of what was happening with some of this turns out to be wrong, off. The on trigger for pull request is needed.

I've added it back to all the workflows to sit beside push. (I think on push, only works on the repo you push it to, so it doesn't show up on the PR if its in a different repo.[ideally I would have presumed this was the default in all this!]). Sorry @cberner my messup!

cberner:
No worries. Thanks!

PR #119, opened 2021-02-26
Sans-io core: Make low-level request API more complete and self-contained

wmanley:
With this change every operation (enum variant of `ll::request::Operation`) is now its own fully-fledged opaque struct. Data is accessed through accessor function.  The usual sync trait API is unchanged, but is now a simple wrapper around the `ll::Request`. This is intended to compliment #112 by separating the bits that don't involve i/o (or blocking) from the bits that do.

The eventual intention is for ll::Request to become an API in its own
right on top of which the traditional sync trait or future async APIs can
be built.  It contains no I/O functions and only operates on buffers
provided to it (the sans-io approach).  Dealing with the enum based `Request`
API provides additional flexibility - particularly WRT threading and async.
The user will eventually be able to pull a chunk of data from the FUSE
device and send it to a rayon threadpool, or an async executor, etc. to
process it.

Every operation corresponds to an opaque struct.  We can make additions
without breaking backwards compatibility.  In particular in the future I'd
like to make it so we can support multiple ABI versions concurrently - and
this API supports that.

There is no public API break intended.

There is more work to be done to bring the vision to completion, but ideally I'd like to merge this work incrementally, rather than making a change in a big bang. I've been careful not to break any APIs, keeping the change entirely internal for now, so that it can be merged rather than getting stuck on a long lived branch :).

The next step is a low-level reply API, but I've not started work on that yet.


src/ll/request.rs
```diff
@@ -43,378 +136,1216 @@ impl fmt::Display for RequestError {
...
+        }
+    }
+    #[derive(Debug)]
+    pub struct FSyncDir<'a>(&'a fuse_fsync_in);
+    impl<'a> FSyncDir<'a> {
+        pub fn file_handle(&self) -> FileHandle {
+            FileHandle(self.0.fh)
+        }
+        pub fn datasync(&self) -> bool {
+            self.0.fsync_flags & 1 != 0
```
cberner, line 643: This should use the new `FUSE_FSYNC_FDATASYNC` constant you added, instead of `1`

src/request.rs
```diff
@@ -133,582 +125,415 @@ impl<'a> Request<'a> {
...
+            ll::Operation::Rename(x) => {
                 se.filesystem.rename(
                     self,
-                    self.request.nodeid(),
-                    &name,
-                    arg.newdir,
-                    &newname,
+                    self.request.nodeid().into(),
+                    x.from().1,
+                    x.to().0.into(),
```
cberner, line 211: I think all these tuples of (inode, name) will be easier to read if you introduce a struct with named fields, like you did for `CopyFileRangeFile`

cberner (approved):
Overall looks good to me! I had a couple minor comments.

@ianoc do you want to take a look as well, since you're working on the async PR?

src/request.rs
```diff
@@ -64,19 +55,20 @@ impl<'a> Request<'a> {
...
         match self.request.operation() {
             // Filesystem initialization
-            ll::Operation::Init { arg } => {
-                let reply: ReplyRaw<fuse_init_out> = self.reply();
+            ll::Operation::Init(x) => {
+                let reply: ReplyRaw<abi::fuse_init_out> = self.reply();
                 // We don't support ABI versions before 7.6
-                if arg.major < 7 || (arg.major == 7 && arg.minor < 6) {
-                    error!("Unsupported FUSE ABI version {}.{}", arg.major, arg.minor);
+                let v = x.version();
```
ianoc, line 61: Could we use tuple deconstruction here, or a wrapper/new type around the version to provide accessors/ordering?

ianoc:
This looks great to me! I think there are some other spots(but maybe the same type) of the `.0` and `.1` accessors that are worth removing, i left a minor other request/comment. But all super minor.

src/request.rs
wmanley: Note: tuples do provide proper ordering, that's one of the advantages of using a tuple here.  See the next line: `v < (7, 6)` which was a big improvement on the previous `arg.major < 7 || (arg.major == 7 && arg.minor < 6)`.

I'll change the assignments below to use tuple deconstruction as you suggest.

wmanley: Hmm, this didn't work:

```rust
                (se.proto_major, se.proto_minor) = v;
```

Failed with:

```
error[E0070]: invalid left-hand side of assignment
  --> src/request.rs:68:50
   |
68 |                 (se.proto_major, se.proto_minor) = v;
   |                 -------------------------------- ^
   |                 |
   |                 cannot assign to this expression
   |
   = note: destructuring assignments are not currently supported
   = note: for more information, see https://github.com/rust-lang/rfcs/issues/372
```

I guess it only works with `let` :(.

wmanley: I added a newtype afterall

src/request.rs
wmanley: Done.  There may be scope to apply `struct FilenameInDir` more widely in the future to operations like `Lookup`, `Unlink`, etc.  I'll have a think about that.

src/ll/request.rs
wmanley: Fixed, good spot.  I could have sworn that I'd already done that :)

cberner:
Merged. Thanks!

PR #120, opened 2021-02-27
Remove caching of the target for now

ianoc:
It seems this is a common/known issue thats being worked on to do with github actions/cache on macOS 

https://github.com/actions-rs/cargo/issues/111

Causing some flakiness it looks like, similar to those bug reports. For some reason serde crate seems to be the issue for a lot of people curiously enough given reports of a tar issue.

### DIFF
--- a/.github/workflows/macos_ci.yml
+++ b/.github/workflows/macos_ci.yml
@@ -30,11 +30,6 @@ jobs:
         with:
           path: ~/.cargo/git
           key: ${{ runner.os }}-cargo-index-${{ hashFiles('**/Cargo.lock') }}
-      - name: Cache cargo build
-        uses: actions/cache@v1
-        with:
-          path: target
-          key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
       - name: Build all targets
         run: cargo build --all --all-targets
       - name: Build all targets, all features


.github/workflows/macos_ci.yml
```diff
@@ -30,11 +30,6 @@ jobs:
         with:
           path: ~/.cargo/git
           key: ${{ runner.os }}-cargo-index-${{ hashFiles('**/Cargo.lock') }}
-      - name: Cache cargo build
-        uses: actions/cache@v1
```
cberner, line 34: Does the `v2` release fix it?

https://github.com/actions/cache

cberner (approved):
Had a question, but otherwise looks good to me

.github/workflows/macos_ci.yml
ianoc: 2.1.4 has the fix, but it seems there's some sort of gtar issue and i believe the v2 tag doesn't point at it -- https://github.com/actions/cache/issues/528 . A few rust repos from the issues look like they have manually pinned to 2.1.4 to get the original issue fixed. So we could probably try that here alternatively, assuming the gtar issue doesn't happen?

cberner: I see. Ok, I'll just merge this for now then

PR #122, opened 2021-02-28
Fix AutoUnmount

### LINKED ISSUE #121: AutoUnmount requires AllowRoot or AllowOther
The AutoUnmount option requires that AllowRoot or AllowOther also be provided, unless the user is root, so that the fusermount3 binary can unmount the filesystem. We should check this during `check_option_conflicts`

cberner:
Previously it only worked if AllowRoot or AllowOther was also set. This was because the fusermount3 binary was unable to automatically unmount these unless they were run as root, since it runs as the root user and so needs access to the FUSE filesystem.

Fixes #121


cberner:
This failure is due to #125

PR #124, opened 2021-02-28
Only run tests once for PRs

cberner:
With 'push' & 'pull_request' all the tests get run twice for each PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,6 @@
 name: CI
 
 on:
-  push:
   pull_request:
 
 jobs:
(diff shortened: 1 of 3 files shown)


cberner:
@ianoc jfyi. Let me know if you know a better way to fix this. Right now everything is getting run twice

ianoc:
@cberner does every push to the PR trigger a new test? or just one extra one the first time the PR is opened?

ianoc:
ugh, i did repo this. And... it seems github has no options to really fix it :/ 

there is some other actions that try do skipping of concurrent runs, but it really gunks up the output from a PR being full of skipped entries. (They point at this config of or in the docs, but alas seems to be endless forums talking about this with no real solutions....). Running twice will just slow it all down and be super noisy. This lgtm

PR #128, opened 2021-03-03
Safety: Make Channel/Session safer by clarifying ownership

wmanley:
This splits `Channel` in two.  `Channel` now only encapsulates reading from the FUSE device.  It is no longer responsible for unmounting the filesystem.  Unmounting falls to a new, backend-specific, struct `Mount`. `Channel` and `ChannelSender` now have shared ownership (via `Arc`) of the FUSE device, and it's closed when it's no longer needed.

The `Mount` belongs to session.  This way the `Session` can drop the mount
without also closing the fuse device file. This is important as we don't
want to close an FD that is still in use as we can cause the equivalent
of use-after-free bugs†.  When a filesystem is unmounted the devices
return `ENODEV`, so should naturally exit as well in an orderly fashion
and without races.

Ownership is now a lot more straightforward.  In particular the libfuse3
`fuse_session` lifecycle was a little confused before, while now we use
RAII in a straightforward fashion to ensure things are cleaned up
correctly, and not cleaned up twice.  The previous implementation had a
bug where it would never be freed because `fuse_session_unmount` would
only be called on null pointers.

The fuse device objects are plain `std::fs::File`s with ownership over the
underlying FD.  Note: although they are Files, we still implement our own
reading and writing via `libc::read` and `libc::writev` due to concerns
about the guarantees that `std::fs::File`'s `AsyncRead` and `AsyncWrite`
provides.  We can still revisit this decison in the future.

Generally I've followed the principle that it's a good idea to encapsulate
an FD as soon as we can for greater confidence that it will get cleaned
up correctly.

This contains API additions, but no API breakage.

**TODO:**

- [x] Re-instate polling fd before attempting unmount

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -18,6 +18,6 @@ task:
     - cargo doc --all --no-deps --features=abi-7-21
   test_script:
     - . $HOME/.cargo/env
-    - cargo test --all --all-targets
-    - cargo test --all --all-targets --features=abi-7-21
+    - cargo test --all --all-targets -- --skip=channel::test::mount_unmount
+    - cargo test --all --all-targets --features=abi-7-21 -- --skip=channel::test::mount_unmount
   before_cache_script: rm -rf $CARGO_HOME/registry/index
--- a/src/channel/fuse_pure.rs
+++ b/src/channel/fuse_pure.rs
@@ -63,9 +63,7 @@ impl Drop for Mount {
             // fusermount in auto-unmount mode, no more work to do.
             return;
         }
-        let rc = super::libc_umount(&self.mountpoint);
-        if rc < 0 {
-            let err = io::Error::last_os_error();
+        if let Err(err) = super::libc_umount(&self.mountpoint) {
             if err.kind() == PermissionDenied {
                 // Linux always returns EPERM for non-root users.  We have to let the
                 // library go through the setuid-root "fusermount -u" to unmount.
--- a/src/channel/fuse_pure.rs
+++ b/src/channel/fuse_pure.rs
@@ -30,14 +30,17 @@ use super::*;
 #[derive(Debug)]
 pub struct Mount {
     mountpoint: CString,
+    auto_unmount_socket: Option<UnixStream>,
 }
 impl Mount {
     pub fn new(mountpoint: &Path, options: &[MountOption]) -> io::Result<(File, Mount)> {
         let mountpoint = mountpoint.canonicalize()?;
+        let (file, sock) = fuse_mount_pure(mountpoint.as_os_str(), options)?;
         Ok((
-            fuse_mount_pure(mountpoint.as_os_str(), options)?,
+            file,
             Mount {
                 mountpoint: CString::new(mountpoint.as_os_str().as_bytes())?,
+                auto_unmount_socket: sock,
             },
         ))
     }
@@ -46,7 +49,11 @@ impl Mount {
 impl Drop for Mount {
     fn drop(&mut self) {
         use std::io::ErrorKind::PermissionDenied;
-
+        if let Some(sock) = mem::take(&mut self.auto_unmount_socket) {
+            drop(sock);
+            // fusermount in auto-unmount mode, no more work to do.
+            return;
+        }
         let rc = super::libc_umount(&self.mountpoint);
         if rc < 0 {
             let err = io::Error::last_os_error();
@@ -61,7 +68,10 @@ impl Drop for Mount {
     }
 }
 
-fn fuse_mount_pure(mountpoint: &OsStr, options: &[MountOption]) -> Result<File, io::Error> {
+fn fuse_mount_pure(
+    mountpoint: &OsStr,
+    options: &[MountOption],
+) -> Result<(File, Option<UnixStream>), io::Error> {
     if options.contains(&MountOption::AutoUnmount) {
         // Auto unmount is only supported via fusermount
         return fuse_mount_fusermount(mountpoint, options);
@@ -69,7 +79,7 @@ fn fuse_mount_pure(mountpoint: &OsStr, options: &[MountOption]) -> Result<File, 
 
     let res = fuse_mount_sys(mountpoint, options)?;
     if let Some(file) = res {
-        Ok(file)
+        Ok((file, None))
     } else {
         // Retry
         fuse_mount_fusermount(mountpoint, options)
@@ -200,7 +210,10 @@ fn receive_fusermount_message(socket: &UnixStream) -> Result<File, Error> {
     }
 }
 
-fn fuse_mount_fusermount(mountpoint: &OsStr, options: &[MountOption]) -> Result<File, Error> {
+fn fuse_mount_fusermount(
+    mountpoint: &OsStr,
+    options: &[MountOption],
+) -> Result<(File, Option<UnixStream>), Error> {
     let (child_socket, receive_socket) = UnixStream::pair()?;
 
     unsafe {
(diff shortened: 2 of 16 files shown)


cberner:
Cool! I'll take a look this weekend, if not sooner

ianoc:
```
This is important as we don't
want to close an FD that is still in use as we can cause the equivalent
of use-after-free bugs†
```
could you expand on this? I'm not sure there are use after free concerns, if the socket is closed then all the calls will fail. If we are concerned about using a re-allocated FD we probably should address that directly. (fwiw, this code and the master code is calling umount, which isn't done in libfuse or the go impl that i can work out without the autounmount flag being present, which we should probably be following too).

wmanley:
>  If we are concerned about using a re-allocated FD we probably should address that directly.

Yes, this is the concern.  For example: you'd be pretty miffed if in your multi-threaded app you open SQLite in one thread while closing your FUSE device and you ended up with a corrupt database due to the race condition.  This is addressed in f3a5499.

src/channel/fuse_pure.rs
```diff
@@ -6,130 +6,87 @@
...
-            let result = libc::poll(&mut poll_result, 1, 0);
-            libc::close(fd);
-            // If the filesystem has already been unmounted, avoid unmounting it again
-            // Unmounting it a second time could cause a race with a newly mounted filesystem
-            // living at the same mountpoint
-            if result > 0 && (poll_result.revents & libc::POLLERR) != 0 {
-                return;
-            }
-        }
-    }
```
cberner, line 131: Why did you remove this? I believe I based this on the behavior of libfuse, which has the same logic

src/session.rs
```diff
@@ -166,31 +180,27 @@ impl BackgroundSession {
...
-        }
+    /// Unmount the filesystem and join the background thread.
+    pub fn join(self) {
+        let Self {
+            mountpoint: _,
+            guard,
+            _mount,
+        } = self;
+        drop(_mount);
+        guard.join().unwrap().unwrap();
```
cberner, line 203: It looks like users need to call `join()` instead of just dropping the `BackgroundSession` now, which will break existing code. Why change it?

cberner (commented):
> The previous implementation had a
bug where it would never be freed because fuse_session_unmount would
only be called on null pointers.

Oops, good catch!

Overall looks good to me. I had a couple questions though

cberner:
> ```
> This is important as we don't
> want to close an FD that is still in use as we can cause the equivalent
> of use-after-free bugs†
> ```
> 
> could you expand on this? I'm not sure there are use after free concerns, if the socket is closed then all the calls will fail. If we are concerned about using a re-allocated FD we probably should address that directly. (fwiw, this code and the master code is calling umount, which isn't done in libfuse or the go impl that i can work out without the autounmount flag being present, which we should probably be following too).

Good point about not unmounting unless `AutoUnmount` is present. The calls to unmount are from the original `fuse-rs` code, but I believe you're right that libfuse doesn't do that, and so we shouldn't either

src/channel/fuse_pure.rs
wmanley: This was a mistake, I'd intended to add it back in after the refactoring to do `Arc`. I'll put it back.

src/session.rs
wmanley: The previous `Drop` impl doesn't call join AFAICS.  I think the previous behaviour was to attempt the unmount on drop.  This change preserves this behaviour.

The `join` method differs from drop in that unlike before we *do* wait for the background thread to complete. Unless I'm misunderstanding?

wmanley:
> Good point about not unmounting unless `AutoUnmount` is present. The calls to unmount are from the original `fuse-rs` code, but I believe you're right that libfuse doesn't do that, and so we shouldn't either

I'm not familiar with libfuse. What's the use-case for having a fuse filesystem mounted that doesn't work? IOW why would not want auto-unmount?

Looking at the man page:

```
   libfuse-specific mount options:
       These following options are not actually passed to the kernel but
       interpreted by libfuse. They can be specified for all filesystems
       that use libfuse:

       auto_unmount
              This option enables automatic release of the mountpoint if
              filesystem terminates for any reason. Normally the
              filesystem is responsible for releasing the mountpoint,
              which means that the mountpoint becomes inaccessible if
              the filesystem terminates without first unmounting.
```

The way I read  "normally the filesystem is responsible for releasing the mountpoint" is that filesystems *should* be calling unmount explicitly, but the `auto_unmount` option is for enabling the trick where an external process is responsible for the unmount when that socket closes. I guess that helps cleanup when a fuse filesystem process is forcibly killed or dies unexpectedly.  I don't think it means that it's desirable to leave a filesystem mounted if it is broken - it just controls the mechanism by which cleanup takes place.

As I say, I'm not familiar with libfuse, so I could easily be missing something here.

As it stands this PR preserves (or attempts to preserve) the existing semantics. If we want to fix/change the handling of automount IMO that deserves its own PR.

src/channel/fuse_pure.rs
wmanley: I've added this back in, but I'm not sure how useful it will be in practice.  I've been unable to get `is_mounted` to return `false` because if it's just detached (lazy unmount) then the `poll` succeeds.

src/session.rs
cberner: Oh yes, my bad, I didn't think this through and thought that without `impl Drop for BackgroundSession` it wouldn't unmount. However, `Drop` is implemented on the `Mount` object now

.cirrus.yml
```diff
@@ -18,6 +18,7 @@ task:
     - cargo doc --all --no-deps --features=abi-7-21
   test_script:
     - . $HOME/.cargo/env
+    - export SKIP_DEVICE_TESTS=1
```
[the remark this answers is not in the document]
wmanley: I've fixed this. Now there's just the issue that `test_mount_unmount` hangs in CI on MacOS. This is the problem with writing tests, unfortunately sometimes you find bugs :crying_cat_face:.  It's strange because the integration tests still pass...

wmanley:
> `test_mount_unmount` hangs in CI on MacOS

I'm not sure what to do here.  I don't have a Mac to debug this on. If someone could tell me where it's blocked (with `strace`) then I can investigate further.

cberner:
I don't use a Mac for development either. The last time I debugged a Mac issue, I added debug code to the Mac CI and print to the console. If you want an interactive session, you could also look at EC2. I think they offer Mac instances now

wmanley:
From my investigation: it appears that unmount fails on MacOS. This causes this test to fail. The test then hangs when attempting to delete the temporary directory becuase it tries to access the mountpoint.

See logs here: https://github.com/cberner/fuser/pull/128/checks?check_run_id=2087641296

From reading the [code to osxfuse](https://github.com/osxfuse/fuse/blob/c6a59bed40bb526c113f861ed49a78b5f224805a/lib/mount_darwin.c) I believe that unmounting on MacOS relies on autounmount - i.e. there's a control socket to the child `mount_macfuse` process and when that is closed the FS is unmounted by the child. Unfortunately from reading the code it seems that this socket FD is leaked.  So it means that the FS will be unmounted when the process dies, but not before.  Because we don't know which socket it is we can't close it ourselves to cause the mount to end before the process does, if the call to `umount` fails.

One possibility would be to pass `MNT_FORCE` to umount. This is what we do in `fuse_unmount_pure` on MacOS. 

What I'll do for now is to fix the test such that it leaks the tempdir on failure, so we don't get a hang.  Also I'll skip the unmount assertion on non-linux as on Linux it should always succeed (thanks to `MNT_DETACH`) while it can fail on other systems.

wmanley:
The test "passes" now, so CI is green.

@cberner: I've rebased on master to resolve conflicts, but really only the last 4 commits are new since last time you reviewed.

cberner:
Great, I'll take a look!

cberner:
@wmanley looks good to me! Could you rebase, there's a conflict? Then I'll merge this

wmanley:
> @wmanley looks good to me! Could you rebase, there's a conflict? Then I'll merge this

Thanks for the review. Rebased.

cberner:
Merged. Thanks!

PR #130, opened 2021-03-10
Api changes to support async api

ianoc:
This is the API boundary core, without an actual impl.

The test updates here cover the reply.rs async version.

Future PR's needed:

1) Make integration tests support multiple impls
2) Add async examples (1 per PR)
2) Add tokio impl, and update examples to use it

While these are inflight building with all flags enabled will see some unused code. But normal build flags should see no unused code warnings.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,9 +23,9 @@ users = "0.11.0"
 page_size = "0.4.2"
 serde = {version = "1.0.102", features = ["std", "derive"], optional = true}
 zerocopy = "0.3"
-async-trait = "0.1.42"
-futures = "0.3.13"
-derivative = "2.2.0"
+async-trait = {version = "0.1.42", optional = true}
+futures = {version = "0.3.13", optional = true}
+derivative = {version = "2.2.0", optional = true}
 
 [dev-dependencies]
 env_logger = "0.8"
@@ -41,7 +41,8 @@ pkg-config = {version = "0.3.14", optional = true }
 default = ["libfuse"]
 libfuse = ["pkg-config"]
 serializable = ["serde"]
-async_impl = []
+async_api = ["async-trait", "futures", "derivative"]
+async_impl = ["async_api"]
 abi-7-9 = []
 abi-7-10 = ["abi-7-9"]
 abi-7-11 = ["abi-7-10"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,12 +23,16 @@ users = "0.11.0"
 page_size = "0.4.2"
 serde = {version = "1.0.102", features = ["std", "derive"], optional = true}
 zerocopy = "0.3"
+async-trait = "0.1.42"
+futures = "0.3.13"
+derivative = "2.2.0"
 
 [dev-dependencies]
 env_logger = "0.8"
 clap = "2.32"
 bincode = "1.3.1"
 serde = {version = "1.0.102", features=["std", "derive"]}
+futures-await-test = "0.3.0"
 
 [build-dependencies]
 pkg-config = {version = "0.3.14", optional = true }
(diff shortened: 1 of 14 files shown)


wmanley:
> The test updates here cover the reply.rs async version.

I've been working on a sans-io `Reply` type which should help reduce duplication here. I've raised my WIP as #131 to show.

src/lib.rs
```diff
@@ -40,6 +40,7 @@ use std::cmp::max;
 #[cfg(feature = "abi-7-13")]
 use std::cmp::min;
 
+pub mod async_api;
```
cberner, line 43: I'm not sure, but I think you can add `#[cfg(feature = "async_impl")]` here, and then you won't need to add it all over the `async_api` module

src/async_api/reply.rs
```diff
@@ -0,0 +1,1336 @@
...
+        nlink: attr.nlink,
+        uid: attr.uid,
+        gid: attr.gid,
+        rdev: attr.rdev,
+        #[cfg(feature = "abi-7-9")]
+        blksize: attr.blksize,
+        #[cfg(feature = "abi-7-9")]
+        padding: attr.padding,
+    }
+}
```
cberner, line 139: Instead of copying these, can you import them from the sync `reply.rs`?

cberner (approved):
Could you add compiling with this feature flag to the CI, and also add `cargo clippy --all-targets --features async_impl` to the `pre` target of the Makefile? Just want to be sure we have some basic test coverage to be sure build errors don't creep into this code, because it's not compiled by default.

I had some other minor comments too. Otherwise it looks good to me though

cberner:
> > The test updates here cover the reply.rs async version.
> 
> I've been working on a sans-io `Reply` type which should help reduce duplication here. I've raised my WIP as #131 to show.

Awesome, that'll be great to cut down on duplication! I'll take a look at that PR next.

Cargo.toml
```diff
@@ -23,12 +23,16 @@ users = "0.11.0"
 page_size = "0.4.2"
 serde = {version = "1.0.102", features = ["std", "derive"], optional = true}
 zerocopy = "0.3"
+async-trait = "0.1.42"
+futures = "0.3.13"
+derivative = "2.2.0"
```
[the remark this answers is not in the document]
ianoc: I can add another flag for that -- right now these are used based on there being the 'async code' vs async impl. The thought here is you can build a crate foo to depend on fuser with default flags, and implement the actual fs using any impl since the API doesn't change -- the final crate/bin/etc could then add the right async impl flag without a conflict and start it with that.

src/async_api/active_session.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// If an init message has been received/processed then this should return Some
+    /// otherwise None.
+    async fn session_configuration(&self) -> Option<SessionConfiguration>;
+}
+
+#[async_trait::async_trait]
+pub(in crate::async_api) trait Worker: Send + Sync {
+    // this method can presumed to only return when the worker/session is being shut down
+    // generally something like a one shot channel will back this.
+    async fn return_on_destory(&mut self) -> ();
```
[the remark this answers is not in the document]
ianoc: sgtm

src/lib.rs
ianoc: i'll add an async_api flag, but i don't think it'll avoid that since building agnostic crates requires a separation between exposed api vs particular implementation i believe

src/async_api/reply.rs
ianoc: 👍

ianoc:
I updated the API function name as you suggested, seems better i think.

I made the dependencies and module optional gated on a `async_api` flag, added new checks around it. and removed the duplication

src/lib.rs
cberner: Ah, I see. I misunderstood why the `cfg`s were there. Either way works for me, so feel free to change it back in a future PR if you want

cberner:
Merged. Thanks!

PR #131, opened 2021-03-10
Sans-IO Reply type: `ll::Response`

wmanley:
Raising for early visibility.

Should help reduce duplication between sync and async APIs, e.g. with #130. 

**TODO**

- [x] Merge #128 and rebase
- [x] Implement the rest of the Reply types:
    - [x] entry
    - [x] attr
    - [x] write
    - [x] statfs
    - [x] create
    - [x] directory


wmanley:
@cberner: I now consider this ready for review. I've not yet deduplicated `async_api/reply.rs` as I have a more extensive deduplication planned for a future PR and I don't want to duplicate that work.

cberner:
Great! I'm in the process of reverting the async changes, since #133 is abandoned, but then will take a look at this

cberner (commented):
Started going through the commits one at a time, since there are so many. Note to myself: this one looks good, "ll::Request: Downcast to operation on-demand"

Also, mind rebasing? There are some conflicts after I removed all the unfinished async code

wmanley:
> Started going through the commits one at a time, since there are so many. 

Yeah, there's a lot.  I tried to break them up to make it easier to review/bisect.

> Also, mind rebasing?

Done

cberner:
> > Started going through the commits one at a time, since there are so many.
> 
> Yeah, there's a lot. I tried to break them up to make it easier to review/bisect.
> 
Yep, definitely appreciated! It's much easier to review this way

src/reply.rs
```diff
@@ -849,12 +849,16 @@ mod test {
...
     struct AssertSender {
-        expected: Vec<Vec<u8>>,
+        expected: Vec<u8>,
     }
 
     impl super::ReplySender for AssertSender {
         fn send(&self, data: &[&[u8]]) {
-            assert_eq!(self.expected, data);
+            let mut v = vec![];
+            for x in bufs {
```
cberner: I think there's a typo here. `bufs` isn't defined. Should be `data` I think?

src/reply.rs
```diff
@@ -853,12 +858,13 @@ mod test {
     }
 
     impl super::ReplySender for AssertSender {
-        fn send(&self, data: &[&[u8]]) {
+        fn send(&self, data: &[IoSlice<'_>]) -> std::io::Result<()> {
             let mut v = vec![];
-            for x in bufs {
+            for x in data {
```
cberner, line 680: Looks like this fixes a typo in the previous commit. Can you fix it there instead?

cberner (commented):
Noticed one typo. Otherwise the commits up through "ReplyRaw: Check for integer overflow" look good.
Nice optimization with `IoSlice` :)

src/ll/reply.rs
```diff
@@ -47,4 +49,34 @@ impl Response {
         }
         f(&v)
     }
+
+    // Constructors
+    pub(crate) fn new_empty() -> Self {
+        Self::Error(0)
```
cberner, line 68: `Data([])` might be more clear than `Error(0)`, since this is not an error

cberner (commented):
Note to self, need to review the following still:

- [x] Reply: Port `ReplyXAttr` to `ll::Response`
- [x] Reply: Port ReplyEntry to `ll:Response`
- [x] Reply: Port `ReplyCreate` to `ll::Response`
- [x] Reply: Port `ReplyIoctl` to `ll::Response`
- [x] Reply: Port `ReplyDirectoryPlus` to `ll::Response`
- [x] op::ListXAttr: Add `reply` function
- [x] ll::Request: Return `Path` rather than `OsStr` where it makes sense
- [x] Reply: Port init reply to `ll::Response`
- [x] Reply: Remove fuse_abi references
- [x] ll::Request: Require `Request` impls be `Sized`

cberner:
@wmanley I think I'm missing some context. There are a bunch of new `.reply()` methods on the types in `ll::request`, but they're all marked as dead code. What are they intended for?

src/reply.rs
wmanley: Fixed in #137.

wmanley:
> Noticed one typo. Otherwise the commits up through "ReplyRaw: Check for integer overflow" look good.

I've raised #137 with the commits up till there so as to allow merging incrementally. I've fixed the typo there.  I'll rebase this PR once that's merged.

wmanley:
> @wmanley I think I'm missing some context. There are a bunch of new `.reply()` methods on the types in `ll::request`, but they're all marked as dead code. What are they intended for?

Yes indeed, they are not used yet.  With this PR and #119 I've been working towards creating #136 - an enum based API which I believe offers significant advantages as compared to a trait based one. You can get a feel for how the `.reply()` will be used looking here: https://github.com/cberner/fuser/pull/136/commits/f05d437f5b3c5aa05ccfb821ad7787401549e8a8?diff=unified&w=1 (view hiding whitespace changes to get a better feel for it). Basically: users can't create `Response`s directly because the `Response::new_*` methods are `pub(crate)`. They instead create them by calling `.reply()` on the operation they've been passed. `op.reply` will always return a `Response` of the correct type for that `Request` type.

`reply` isn't used in this PR, but it seemed natural to include it here as I was working through each reply type.

cberner:
Ah, I see. Cool, I'll take a look at that. I merged #137 :)

wmanley:
Rebased since #137 is merged

wmanley:
I just fixed some clippy lints and have rebased them into the relevant commits. No exciting changes made.

cberner:
Ok, now that I understand how #136 works I think I see what these are for. I'd like to merge this work in a way that doesn't involve adding dead code though. Is there an obvious way to break it up into medium sized PRs that add end-to-end functionality?

src/ll/reply.rs
```diff
@@ -316,8 +316,58 @@ impl From<crate::FileAttr> for Attr {
     }
 }
 
+#[derive(Debug)]
+pub struct EntListBuf {
```
wmanley, line 323: Hmm, this shouldn't be pub.

wmanley:
> Ok, now that I understand how #136 works I think I see what these are for. I'd like to merge this work in a way that doesn't involve adding dead code though. Is there an obvious way to break it up into medium sized PRs that add end-to-end functionality?

Unfortunately I don't know how to do this. I'm trying to gradually work towards #136, but until we make these types public there is going to be code in this crate that is unused, because it's not yet callable from outside the crate. I see the `#[allow(dead_code)]` as an advantage really - if this work does come to naught then it's easy to find the code that can be deleted by grepping for that.

Two alternatives:

* Make these APIs part of the crate's public interface now, rather than later.
* Don't add the `reply` functions in this PR, but instead merge them all in big PR later along with the #136 and any other code that is required for the enum API.

I think it's a bit premature for the former, and the latter seems like it would be hard to review.

cberner:
I see, ok I'll review this as-is then! Should have time this weekend. Maybe sooner.

One thing, I may end up asking for changes to these pieces, when we get to the future PRs and I see how everything fits together. Just fair warning :)

src/ll/request.rs
```diff
@@ -907,6 +907,24 @@ mod op {
         pub fn size(&self) -> u32 {
             self.arg.size
         }
+        #[allow(dead_code)]
+        pub fn reply<It: Iterator<Item = T>, T: AsRef<OsStr>>(&self, it: It) -> Response {
+            if self.size() == 0 {
+                // Probing to find an appropriate size for the buffer
+                let size: usize = it.map(|x| x.as_ref().len() + 1).sum();
+                Response::new_xattr_size(size.try_into().expect("To many XAttrs"))
```
cberner, line 914: typo in except message. Should be "Too many"

cberner (approved):
Looks good to me! I had just one comment about a typo I noticed

src/ll/reply.rs
wmanley: I did it this way as it will be mildly more efficient - we don't need to faf around with `SmallVec` in this case.

cberner:
Merged. Thanks!

PR #132, opened 2021-03-11
Add in the support in the tests for doing async/tokio tests

ianoc:
This hopefully is a relatively boring PR. We may want to do more cleanups later too.

But:

Moves some setup in the docker stuff from the build to the run, pass in the args via the github config of what to run.

Motivation:

its only a small change from here to enable integration testing for tokio. But since that PR will be bigger figure it would be nice to separate this one out.


ianoc:
Updated per comments to ensure saner bash

cberner:
Merged. Thanks!

PR #137, opened 2021-03-18
Prep commits from #131

wmanley:
These were reviewed by @cberner as part of #131.  I've fixed the issue with the bufs typo in "Refactor reply tests: Join expected data together" and rebased.

I'm raising this to allow this work to be merged incrementally so as to make any review and rework of #131 easier.


cberner:
Merged. Thanks!

PR #139, opened 2021-04-05
Refactor: Isolate mount backend specific code

wmanley:
There is now no `[cfg(feature="libfuse")]`... code outside of `mod mnt`.


cberner:
Merged. Thanks!

PR #140, opened 2021-04-06
Various ll::request improvements

wmanley:
Made while working on porting simple.rs to the enum API and in preparation for #136


cberner:
Merged. Thanks!

PR #141, opened 2021-05-02
Update readme with homebrew installation instructions

orf:
`brew cask ...` is no longer a valid command, and the 4.x version is called `macfuse` now.


cberner:
Merged. Thanks!

PR #154, opened 2021-06-18
Ensure that Filesystem.destroy() is always called

### LINKED ISSUE #153: destroy() not being called on Linux
I'm using fuser on macOS and Linux. My filesystem writes data back out when `destroy()` is called, which works fine on macOS. On Linux, it seems like `destroy()` is never getting called.

This issue may be old: https://github.com/zargony/fuse-rs/issues/151. Others have had issues with libfuse not properly calling destroy due to [signal handlers](https://github.com/libfuse/libfuse/issues/228).

You can see a [successful run on macOS](https://github.com/mgree/ffs/runs/2843591862?check_suite_focus=true#step:5:185); the relevant ["calling sync" lines](https://github.com/mgree/ffs/blob/ec7e63e129ad1a41ee37e4b9a54178747130d60b/src/fs.rs#L185-L189) don't appear [on Linux](https://github.com/mgree/ffs/runs/2843591890?check_suite_focus=true#step:5:40).

I had suspected that calling `fusermount -u` instead of `umount` was the key, but it [makes no difference](https://github.com/mgree/ffs/runs/2843898103?check_suite_focus=true#step:5:280). Killing the process [doesn't help either](https://github.com/mgree/ffs/runs/2843928635?check_suite_focus=true#step:5:278).

cberner:
libfuse uses the same workaround of calling .destroy() if the kernel
never delivers the DESTROY message

Fixes #153

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -34,7 +34,7 @@ pub struct Request<'a> {
 
 impl<'a> Request<'a> {
     /// Create a new request from the given data
-    pub fn new(ch: ChannelSender, data: &'a [u8]) -> Option<Request<'a>> {
+    pub(crate) fn new(ch: ChannelSender, data: &'a [u8]) -> Option<Request<'a>> {
         let request = match ll::AnyRequest::try_from(data) {
             Ok(request) => request,
             Err(err) => {
@@ -49,7 +49,7 @@ impl<'a> Request<'a> {
     /// Dispatch request to the given filesystem.
     /// This calls the appropriate filesystem operation method for the
     /// request and sends back the returned reply to the kernel
-    pub fn dispatch<FS: Filesystem>(&self, se: &mut Session<FS>) {
+    pub(crate) fn dispatch<FS: Filesystem>(&self, se: &mut Session<FS>) {
         debug!("{}", self.request);
         let unique = self.request.unique();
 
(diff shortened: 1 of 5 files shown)


boulos:
This is also a breaking API change to destroy, right? Can't you manufacture a Request instead?

src/request.rs
```diff
@@ -34,7 +34,7 @@ pub struct Request<'a> {
 
 impl<'a> Request<'a> {
     /// Create a new request from the given data
-    pub fn new(ch: ChannelSender, data: &'a [u8]) -> Option<Request<'a>> {
+    pub (in crate) fn new(ch: ChannelSender, data: &'a [u8]) -> Option<Request<'a>> {
```
wfraser, line 37: can be simplified to `pub(crate)`

cberner: Ah thanks! I didn't know that worked

cberner:
> This is also a breaking API change to destroy, right? Can't you manufacture a Request instead?

It is yes, but updating should be easy for people. I don't like the idea of manufacturing a request, because there would be no way to tell whether the data in the request was real then

boulos:
That's fine. Maybe note it in your changelog as well.

>

rarensu:
I don't understand this issue. Surely, when the run() loop comes to an end, Session gets dropped? Why would being in a thread change that? 

Perhaps there is a hidden issue like, due to a panic, instead of the session dropping first because its thread ended, the drops are getting called in a random order.

PR #177, opened 2021-09-19
Add pure rust mount on Macos

kitterion:
Thankfully, only the kernel module part of macfuse is closed source. The [libfuse](https://github.com/osxfuse/fuse) part is still open source, which allowed to do this relatively easily. 
I do want to note, this is not exactly what libfuse does at this point in time, I did not accommodate the changes introduced in [this commit](https://github.com/osxfuse/fuse/commit/c6a59bed40bb526c113f861ed49a78b5f224805a). This should still work for reasons of backwards compatibility of libfuse itself.

### DIFF
--- a/src/mnt/fuse_pure.rs
+++ b/src/mnt/fuse_pure.rs
@@ -130,8 +130,16 @@ fn fuse_unmount_pure(mountpoint: &CStr) {
         .arg(OsStr::new(&mountpoint.to_string_lossy().into_owned()));
 
     if let Ok(output) = builder.output() {
-        debug!("{}: {}", FUSE_PROGRAM_NAME, String::from_utf8_lossy(&output.stdout));
-        debug!("{}: {}", FUSE_PROGRAM_NAME, String::from_utf8_lossy(&output.stderr));
+        debug!(
+            "{}: {}",
+            FUSE_PROGRAM_NAME,
+            String::from_utf8_lossy(&output.stdout)
+        );
+        debug!(
+            "{}: {}",
+            FUSE_PROGRAM_NAME,
+            String::from_utf8_lossy(&output.stderr)
+        );
     }
 }
 
@@ -158,7 +166,7 @@ fn detect_fuse_program() -> String {
 
 #[cfg(target_os = "macos")]
 fn detect_fuse_program() -> String {
-    return "/Library/Filesystems/macfuse.fs/Contents/Resources/mount_macfuse".to_string()
+    "/Library/Filesystems/macfuse.fs/Contents/Resources/mount_macfuse".to_string()
 }
 
 fn receive_fuse_message(socket: &UnixStream) -> Result<File, Error> {
@@ -291,7 +299,11 @@ fn fuse_mount(
             }
             let mut buf = vec![0; 64 * 1024];
             if let Ok(len) = stdout.read(&mut buf) {
-                debug!("{}: {}", FUSE_PROGRAM_NAME, String::from_utf8_lossy(&buf[..len]));
+                debug!(
+                    "{}: {}",
+                    FUSE_PROGRAM_NAME,
+                    String::from_utf8_lossy(&buf[..len])
+                );
             }
         }
         if let Some(mut stderr) = fuse_child.stderr {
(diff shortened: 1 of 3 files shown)


src/mnt/fuse_pure.rs
```diff
@@ -23,9 +23,11 @@ use std::process::{Command, Stdio};
 use std::sync::Arc;
 use std::{mem, ptr};
 
-const FUSERMOUNT_BIN: &str = "fusermount";
-const FUSERMOUNT3_BIN: &str = "fusermount3";
-const FUSERMOUNT_COMM_ENV: &str = "_FUSE_COMMFD";
+#[cfg(target_os = "linux")]
```
kitterion, line 26: Because of all the conditional compilation, there's now a fair bit of dead code/unused functions warnings in this file. I wonder if it's viable to just silence them with `#![allow(dead_code)]` and similar?

src/mnt/fuse_pure.rs
```diff
@@ -138,10 +161,15 @@ fn detect_fusermount_bin() -> String {
...
     }
     // Default to fusermount3
-    FUSERMOUNT3_BIN.to_string()
+    fusermount3_bin.to_string()
 }
 
-fn receive_fusermount_message(socket: &UnixStream) -> Result<File, Error> {
+#[cfg(target_os = "macos")]
+fn detect_fuse_program() -> String {
+    "/Library/Filesystems/macfuse.fs/Contents/Resources/mount_macfuse".to_string()
```
cberner, line 169: On Linux we first search the PATH for `fusermount` or `fusermount3`. Is `mount_macfuse` in the PATH, and should we first use that?

src/mnt/fuse_pure.rs
```diff
@@ -23,9 +23,11 @@ use std::process::{Command, Stdio};
...
 use std::{mem, ptr};
 
-const FUSERMOUNT_BIN: &str = "fusermount";
-const FUSERMOUNT3_BIN: &str = "fusermount3";
-const FUSERMOUNT_COMM_ENV: &str = "_FUSE_COMMFD";
+#[cfg(target_os = "linux")]
+const FUSE_PROGRAM_NAME: &str = "fusermount";
+
+#[cfg(target_os = "macos")]
+const FUSE_PROGRAM_NAME: &str = "macfuse";
```
cberner, line 30: I think this should be `mount_macfuse` based on the binary used in `detect_fuse_program`

cberner (changes requested):
Can you add tests for building without libfuse to `osx_moun_tests.sh`? That way this will get tested in the CI. It should be as simple as adding the lines:
```
run_test --no-default-features 'without libfuse'
run_test --no-default-features 'without libfuse' --auto_unmount
```

src/mnt/fuse_pure.rs
kitterion: It is not on PATH unfortunately, and this is what libfuse does anyway. It takes the [full path](https://github.com/osxfuse/fuse/blob/8071bd97494cd2982ed5635525b8a754edfbfb0f/lib/fuse_darwin_private.h#L41) and [calls the fuse_resource_path function](https://github.com/osxfuse/fuse/blob/8071bd97494cd2982ed5635525b8a754edfbfb0f/lib/mount_darwin.c#L428), which in turn just [returns it as is](https://github.com/osxfuse/fuse/blob/8071bd97494cd2982ed5635525b8a754edfbfb0f/lib/fuse_darwin.c#L323-L327), since it's an absolute path.

cberner: I see, ok. Sounds good then!

boulos:
Does the older osxfuse have the same setup?

kitterion:
> Does the older osxfuse have the same setup?

I looked into it and no, they don't have the same setup, specifically the program name and the env variables are different.
Why, is there a need to support osxfuse?

boulos:
If easy to support with just a "oh, the path is osxfuse in the old one",
then it'd be worthwhile to cover more setups. Otherwise, meh.

>

kitterion:
Regarding the tests, I'm not going to add them just yet, turns out they don't quite work with macfuse. When running tests locally they just hang, with or without my changes. Well it's not exactly the tests, it's the `mount` command that hangs with the only solution being to reboot the system.

cberner:
> Regarding the tests, I'm not going to add them just yet, turns out they don't quite work with macfuse. When running tests locally they just hang, with or without my changes. Well it's not exactly the tests, it's the `mount` command that hangs with the only solution being to reboot the system.

Interesting. Have you tried with `osxfuse`, if that's easy to make work? The current CI passes and it links against `osxfuse` & uses the `osxfuse` mount binary. It'd be nice to have tests for this functionality, so that it doesn't get broken in the future

cberner:
I wasn't able to get this PR working, but wrote CI integration here: https://github.com/cberner/fuser/pull/192

Feel free to re-open if you figure out how to get this working

PR #180, opened 2021-09-26
Warning and comment to clarify AutoUnmount behavior

mgree:
I was caught out by a change in how options are treated in 0.9.1: if you set `AutoUnmount` and neither `AllowRoot` or `AllowOther` are set, then `AllowOther` is getting automatically added. If your `/etc/fuse.conf` doesn't permit `AllowOther`, the mount will fail without much information. See the latter half of #153, after it was closed.

This PR adds a warning when `AllowOther` is added and a comment to the options datatype, in the hopes of saving others some trouble.

### DIFF
--- a/src/mnt/mount_options.rs
+++ b/src/mnt/mount_options.rs
@@ -22,6 +22,8 @@ pub enum MountOption {
     /// Allow the root user to access this filesystem, in addition to the user who mounted it
     AllowRoot,
     /// Automatically unmount when the mounting process exits
+    /// 
+    /// `AutoUnmount` requires `AllowOther` or `AllowRoot`. If `AutoUnmount` is set and neither `Allow...` is set, then `AllowOther` will be automatically added to the options. If the FUSE configuration doesn't permit `AllowOther`, this will cause an `EFAULT` error and mounting will fail.
     AutoUnmount,
     /// Enable permission checking in the kernel
     DefaultPermissions,
--- a/src/session.rs
+++ b/src/session.rs
@@ -6,7 +6,7 @@
 //! for filesystem operations under its mount point.
 
 use libc::{EAGAIN, EINTR, ENODEV, ENOENT};
-use log::info;
+use log::{info,warn};
 use std::fmt;
 use std::path::{Path, PathBuf};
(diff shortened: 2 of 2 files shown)


cberner:
Could you run the formatter to resolve that CI failure?

mgree:
Oops! Sorry about that. 🤭

src/mnt/mount_options.rs
```diff
@@ -22,6 +22,8 @@ pub enum MountOption {
     /// Allow the root user to access this filesystem, in addition to the user who mounted it
     AllowRoot,
     /// Automatically unmount when the mounting process exits
+    ///
+    /// `AutoUnmount` requires `AllowOther` or `AllowRoot`. If `AutoUnmount` is set and neither `Allow...` is set, then `AllowOther` will be automatically added to the options. If the FUSE configuration doesn't permit `AllowOther`, this will cause an `EFAULT` error and mounting will fail.
```
cberner, line 26: Actually, it doesn't automatically add `AllowOther` exactly. It does set `allow_other`, but it enforces permission checks in the library so that other users can't access the mount. Can you change it to something like:
```
If `AutoUnmount` is set and neither `Allow...` is set, the FUSE configuration must permit `allow_other`, otherwise mounting will fail
```

src/session.rs
```diff
@@ -75,6 +75,7 @@ impl<FS: Filesystem> Session<FS> {
             && !(options.contains(&MountOption::AllowRoot)
                 || options.contains(&MountOption::AllowOther))
         {
+            warn!("Given auto_unmount without allow_root or allow_other; adding allow_other");
```
cberner, line 78: can you add `adding allow_other, with userspace permission handling`, or something to that effect?

cberner (changes requested):
Couple clarifications to the message. Otherwise it looks good to me!

src/mnt/mount_options.rs
mgree: Fixed! Sorry you're effectively the one writing the PR now... 🤭

src/session.rs
mgree: Fixed!

PR #182, opened 2021-10-08
Add unimplemented method logging for debugging

### LINKED ISSUE #181: Add logging to default Filesystem method implementations
Currently the majority of `Filesystem` methods have a default implementation which simply return an error. That is, of course, desirable behavior as expecting users to implement all these methods from the get-go is silly. However one issue this creates is that when progressively adding new features it can be unclear which unimplemented methods are causing things to fail.

The options currently are:
* Guess and check
* Make copies of all the unimplemented methods with println statements
* Just keep implementing stuff till it works or fails in a new way

I feel the default implementation should log warnings for each unimplemented method which runs, excluding ones which do not return an error by default (for example `init` does not need this as it is truly optional, while `getxattr` should emit a warning before returning `ENOSYS`)

If this such changes are deemed exceptable I'm more than willing to implement/test/PR such changes, just let me know. :)

jam1garner:
Screenshot example taken from starting up the null example and running `touch [mountpoint]`:

image

Methods which warn (basically the 'hello' example's methods):

* read
* readdir
* getattr
* lookup

Methods which do not log (any which have a default implementation which does not reply with an error):

* init
* destroy
* forget
* batch_forget
* open
* release
* opendir
* releasedir
* statfs

Methods which use 'debug' logging level (basically any for modification, are extensions, or otherwise aren't necessary for a minimal read-only filesystem):

<details>
  <summary>The rest (It's a lot)</summary>

* setattr
* readlink
* mknod
* mkdir
* rmdir
* symlink
* rename
* link
* write
* flush
* fsync
* readdirplus
* fsyncdir
* setxattr
* getxattr
* listxattr
* removexattr
* access
* create
* getlk
* setlk
* bmap
* ioctl
* fallocate
* lseek
* copy_file_range
* setvolname
* exchange
* getxtimesm

</details>

Closes #181 . cc @cberner for review.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,6 +7,7 @@
 #![warn(missing_docs, missing_debug_implementations, rust_2018_idioms)]
 
 use libc::{c_int, ENOSYS};
+use log::{debug, warn};
 use mnt::mount_options::parse_options_from_args;
 #[cfg(feature = "serializable")]
 use serde::{Deserialize, Serialize};
@@ -25,7 +26,6 @@ use crate::mnt::mount_options::check_option_conflicts;
 use crate::session::MAX_WRITE_SIZE;
 #[cfg(feature = "abi-7-16")]
 pub use ll::fuse_abi::fuse_forget_one;
-use log::{debug, warn};
 pub use mnt::mount_options::MountOption;
 #[cfg(target_os = "macos")]
 pub use reply::ReplyXTimes;
@@ -350,7 +350,8 @@ pub trait Filesystem {
         reply: ReplyAttr,
     ) {
         debug!(
-            "[Not Implemented] setattr(ino: {:#x?}, mode: {:?}, uid: {:?}, gid: {:?}, size: {:?}, fh: {:?}, flags: {:?})",
+            "[Not Implemented] setattr(ino: {:#x?}, mode: {:?}, uid: {:?}, \
+            gid: {:?}, size: {:?}, fh: {:?}, flags: {:?})",
             ino, mode, uid, gid, size, fh, flags
         );
         reply.error(ENOSYS);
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -25,6 +25,7 @@ use crate::mnt::mount_options::check_option_conflicts;
 use crate::session::MAX_WRITE_SIZE;
 #[cfg(feature = "abi-7-16")]
 pub use ll::fuse_abi::fuse_forget_one;
+use log::{debug, warn};
```
boulos, line 28: This should be up near the use libc:: include up top.

src/lib.rs
```diff
@@ -356,26 +367,34 @@ pub trait Filesystem {
...
+        mode: u32,
+        umask: u32,
         reply: ReplyEntry,
     ) {
+        debug!(
+            "[Not Implemented] mkdir(parent: {:#x?}, name: {:?}, mode: {}, umask: {:#x?})",
+            parent, name, mode, umask
+        );
         reply.error(ENOSYS);
     }
```
boulos, line 400: Why not debug! on unlink?

src/lib.rs
```diff
@@ -478,15 +517,19 @@ pub trait Filesystem {
...
         _data: &[u8],
-        _write_flags: u32,
-        _flags: i32,
-        _lock_owner: Option<u64>,
+        write_flags: u32,
+        flags: i32,
+        lock_owner: Option<u64>,
         reply: ReplyWrite,
     ) {
+        debug!(
```
boulos, line 529: Can you also log data.len()?

boulos (commented):
This mostly LGTM, but why not add debug! for unlink()? (unlink returns ENOSYS, so you kind of have to do it).

Similarly, statfs has *a* default, but it's not "right". Should that have a debug()! just to tell you that the default is doing so?

Also, did you run `cargo fmt` on this? (The debug! line for copy_file_range is super long, though maybe it's not wrapping because it's a log line, so it can be arbitrarily long)

src/lib.rs
jam1garner: because I am unfortunately a fallible human being!

cberner:
Merged. Thanks!

PR #193, opened 2022-01-03
Use EPERM instead of ENOSYS for link and symlink

madmo:
userspace programs expect EPERM if (sym)linking is not supported.
See man 2 link and man 2 symlink.


cberner:
Thanks!

PR #194, opened 2022-01-11
Update zerocopy to 0.6, clap to 3.0

messense:


messense:
Should I add `BSD-2-Clause` to license allow list?

cberner:
Yes, BSD-2-Clause is fine, go ahead and add it

cberner:
Merged. Thanks!

PR #199, opened 2022-03-03
Added background thread spawn equivalent for mount2.

ph0llux:
This is nothing special, but it would be nice if there is the corresponding background equivalent to mount2 also in the upstream of fuser. I use fuser in several projects and prefer to have something like this directly in the upstream.


cberner:
Cool, this makes sense to me. Could you run `rust fmt` and force push?

ph0llux:
Okay, I ran fmt over it and force pushed. Is there anything missing or does it fit now? ;)

cberner:
Merged. Thanks!

PR #202, opened 2022-03-08
Stop using deprecated clap struct

jvasile:
The compiler was complaining about a deprecated struct and recommended using Command instead of App.  This does that and makes no substantive changes.


cberner:
Merged. Thanks!

PR #207, opened 2022-06-06
Implement Sync and Unpin for ReplySender

Absolucy:
This does not seem to interfere with existing code. The one thing that did mess up was a test that used the std Sender, which is not `Sync`, so I replaced it with a `flume::Sender`.

This is useful for me as I'm implementing a cross-platform async userspace filesystem wrapper (using fuse and dokany), and well, these two traits are kinda needed for async.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,12 +44,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
-name = "bumpalo"
-version = "3.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


cberner:
I'm not very familiar with `Unpin`. Can you give an example of how you're using the fuser crate that requires this? Since `ReplySender` is private, I don't quite understand why this is needed

Absolucy:
Ah, I just assume `Unpin` is useful for async. I can remove it if needed, but it's automatically implemented on nearly every base type anyways.

Also, rustc will automatically implement `Sync` and such for `ReplyRaw` if `ReplySender`, allowing *all* the `Reply*` structs to be `Sync`.

cberner:
I see. That seems fine. I'd rather not add the `flume` dependency though, and it has a dependency that is failing the license checker. Can you remove that? I think `mpsc::sync_channel` might work

Absolucy:
Oh, I didn't know about `sync_channel`, whoops. It's been changed to use that,

cberner:
Merged. Thanks!

PR #210, opened 2022-06-14
Fix CI for MacOS

dev-jwel:
* `osxfuse` has removed as it has been succeeded by `macfuse`

* CI runs for all MacOS versions

* `load_macfuse` is not used as is not required now

* Add timeout for `mnt::test::mount_umount` to check infinity loop


dev-jwel:
143ac80 can pass CI. I tested it at https://github.com/dev-jwel/fuser/actions/runs/2499357942. But it failed for MacOS CI. This is intended result.

PR #218, opened 2022-08-21
Use `ensure_last_os_error()` with libfuse2

str4d:
Under some conditions, `fuse_mount_compat25` may return -1 to indicate
an error but does not set errno. `io::Error::last_os_error()` will then
be equivalent to `Err(Success)` which is somewhat surprising.

In particular, if the mount options `auto_unmount,allow_{root,other}`
are set but the user does not have `user_allow_other` set in their
`fuse.conf`, the `fusermount` binary will print an error message to
stderr but not set errno.

`fusermount3` has the same behaviour, but this issue was mitigated for
the libfuse3 binding in cberner/fuser#178 by mapping `Err(Success)` to
the slightly more useful `Error(Other, "Unspecified Error")`. This
commit applies the same fix to the libfuse2 binding.

### DIFF
--- a/src/mnt/fuse2.rs
+++ b/src/mnt/fuse2.rs
@@ -1,4 +1,4 @@
-use super::{fuse2_sys::*, with_fuse_args, MountOption};
+use super::{ensure_last_os_error, fuse2_sys::*, with_fuse_args, MountOption};
 use log::warn;
 use std::{
     ffi::CString,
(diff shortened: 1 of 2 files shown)


cberner (approved):

str4d:
The `CI / ci (pre)` step is failing because of an unrelated change to the license specifier in `unicode-ident 1.0.3` (https://github.com/dtolnay/unicode-ident/pull/9) to correctly note that anything generated from the Unicode Character Database needs to be used under Unicode-DFS-2016.

cberner:
I merged a fix for that. Could you rebase on master and force push your branch to restart the CI?

str4d:
Rebased on `master` to fix CI. It won't automatically restart though because I still count as a new contributor, so someone needs to check I didn't insert a cryptominer during the rebase 🙃

cberner:
Merged. Thanks!

PR #223, opened 2022-11-18
Make the mount field public

LevitatingBusinessMan:
This allows for unmounting a session running in a different thread, similar to `BackgroundSession`s behavior.

One issue with using `BackgroundSession` is that it require the filesystem to be bound by `Send`.

You can see the benefit [Here](https://github.com/LevitatingBusinessMan/reddit-fs/blob/master/src/main.rs#L447-L471).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,20 +1,18 @@
 [package]
-name = "fuser"
+name = "levitating-fuser"
 edition = "2018"
 version = "0.11.1"
 authors = ["Christopher Berner <christopherberner@gmail.com>"]
-description = "Filesystem in Userspace (FUSE) for Rust"
-documentation = "https://docs.rs/fuser"
-homepage = "https://github.com/cberner/fuser"
-repository = "https://github.com/cberner/fuser"
+description = "My fork of Filesystem in Userspace (FUSE) for Rust"
+repository = "https://github.com/LevitatingBusinessMan/fuser"
 readme = "README.md"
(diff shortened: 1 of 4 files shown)


cberner:
I don't want to expose that field. But I think you can call [umount](https://docs.rs/libc/latest/libc/fn.umount.html) from your ctrl-c handler and it will have the same effect

LevitatingBusinessMan:
Good solution. That works, thanks.

LevitatingBusinessMan:
Heh, on closer inspection it actually doesn't work. It returns `-1`. 

It would be nice if you could add a way to automatically unmount the filesystem (which can't be sent across threads) without having to expose any private fields.

cberner:
@LevitatingBusinessMan can you give this branch a try? https://github.com/cberner/fuser/pull/224

LevitatingBusinessMan:
Looks very good, thanks for taking the time to work on this. I will try to find the time to test it tomorrow.

LevitatingBusinessMan:
Hey, I forgot to test it in the weekend but I tested it now. It works like a charm.

See [here](https://github.com/LevitatingBusinessMan/reddit-fs/blob/master/src/main.rs#L447-L470)

LevitatingBusinessMan:
When will you publish the change to crates, then I can delete my fork from the repositories :)

cberner:
I just released version 0.12.0

PR #225, opened 2022-11-30
Support fuse-t on MacOS

maxburke:
Fuse-t is a user-space FUSE implementation on MacOS; previous versions of FUSE (macfuse/osxfuse) require kernel extensions that are getting progressively more difficult to install as MacOS security tightens. This change adds support for building against `fuse-t`

### DIFF
--- a/.github/workflows/macos_ci.yml
+++ b/.github/workflows/macos_ci.yml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install packages
         run: |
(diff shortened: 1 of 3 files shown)


cberner:
Cool. This seems like a good change. Could you update the macos_ci.yml to use fuse-t instead of osxfuse?

maxburke:
I think I've done what's needed? I don't have much experience with GitHub Actions...

rail-ka:
I just want to point out that not all of the source code of the `fuse-t` project is open source, and does its license affect its use in this project?

https://github.com/macos-fuse-t/fuse-t/issues/1

PR #231, opened 2023-02-23
Use checked arithmetic for addition to Unix epoch.

trombonehero:
The current code can panic with filesystems that do not store times. Using checked arithmetic allows such filesystems to operate without panic.

### DIFF
--- a/src/ll/request.rs
+++ b/src/ll/request.rs
@@ -413,9 +413,8 @@ mod op {
             #[cfg(target_os = "macos")]
             match self.arg.valid & FATTR_CRTIME {
                 0 => None,
-                _ => Some(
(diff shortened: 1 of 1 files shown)


cberner:
I'm not sure this is the right fix. That code path is only taken when the appropriate `FATTR_CRTIME`...etc flag is set. Do you know why that flag is getting set without a valid timestamp passed in?

FirelightFlagboy:
Hello, I'm having an issue that would be solved by this PR.

I've done some investigation and I think the issue lies on `macFuse` side: macfuse/macfuse#1042

@cberner, could this PR be merged ? Or does it require additional work ?

cberner:
Going to close this in favor of https://github.com/cberner/fuser/pull/322

PR #232, opened 2023-03-03
Avoid copy of data when sending Response. 

mgautierfr:
The `Response` structure is owning the data. So it force us to copy (own) the data even if we don't want (nor need) to.
This is especially usefull to avoid copy of the data in ReplyData which can be pretty "big".

This PR introduce a ResponseTrait with a single method returning e `ResponseData` which borrow the data.
It also make `[u8]` and `Response` implement `ResponseTrait`.
So we can now directly pass `&[u8]` (if we don't want to own the data) or still use `Response` if we need to (to pass response from a method to another).

Methods `Response::new_<foo>` have been keep to avoid to much change in the code (and direct creation of `abi::fuse_<foo>_out`). But they may me better moved as free function or `new` method in the struct themselves (or simply removed ?)

### DIFF
--- a/src/ll/reply.rs
+++ b/src/ll/reply.rs
@@ -26,7 +26,7 @@ pub(crate) trait ResponseTrait {
     fn get<'a>(&'a self) -> ResponseData<'a>;
 }
 
-pub(crate) fn send_with_iovec<R: ResponseTrait, F: FnOnce(&[IoSlice<'_>]) -> T, T>(
+pub(crate) fn send_with_iovec<R: ResponseTrait + ?Sized, F: FnOnce(&[IoSlice<'_>]) -> T, T>(
     response: &R,
     unique: RequestId,
     f: F,
--- a/src/ll/reply.rs
+++ b/src/ll/reply.rs
@@ -70,6 +70,12 @@ impl ResponseTrait for Response {
     }
 }
 
+impl ResponseTrait for [u8] {
+    fn get<'a>(&'a self) -> ResponseData<'a> {
+        ResponseData::Data(self)
+    }
+}
+
 #[must_use]
 impl Response {
     // Constructors
--- a/src/ll/reply.rs
+++ b/src/ll/reply.rs
@@ -76,6 +76,7 @@ impl ResponseTrait for [u8] {
     }
 }
 
+
 #[must_use]
 impl Response {
     // Constructors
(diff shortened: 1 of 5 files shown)


src/ll/reply.rs
```diff
@@ -17,57 +17,77 @@ use zerocopy::AsBytes;
...
 pub(crate) type ResponseBuf = SmallVec<[u8; INLINE_DATA_THRESHOLD]>;
 
+pub(crate) enum ResponseData<'a> {
+    Error(i32),
+    Data(&'a [u8])
+}
+
+pub(crate) trait ResponseTrait {
+    fn get<'a>(&'a self) -> ResponseData<'a>;
+}
```
cberner, line 27: Having `ResponseData`, `ResponseTrait`, `ResponseBuf`, and `Response` is more complex than I would like. Is it possible to add something like a `DataRef(&[u8])` to `Response` instead of this additional trait and enum?

cberner (changes requested):

src/ll/reply.rs
mgautierfr: The problem with adding a `DataRef(&[u8])` is that it introduces a lifetime in `Response`. It makes things more complex when `Response` is used as a "owner" of the data (as in [dispatch_req](https://github.com/cberner/fuser/blob/c05bea5817fe201467a2546a4be3718bb60b46f3/src/request.rs#L68))

Another way to simplify this is to move the constrution of the `abi::fuse_*_out` in the structure themselves (with a `new`) and implement `ResponseTrait` for `ResponseBuf`. This way we could simply remove `Response` (and rename `ResponseTrait` to `Response` ?)

jamesbornholt: I had a version of this change that added the `&[u8]` variant to `Response`, and it wasn't so bad: https://github.com/jamesbornholt/fuser/commit/51ca95dd4e469b9e16ce7588b1678570ff1dd03b. I was only trying to optimize the copy out of `ReplyData`, though—I didn't look at whether it was possible to avoid copying other reply types.

cberner: Ya, something like that looks cleaner to me

mgautierfr: It indeed works and is simpler. Thanks for the other PR and have merged it.

PR #233, opened 2023-03-12
Added DeferPermissions option

freewilll:
This is needed for [redox](https://www.redox-os.org/) ([source](https://gitlab.redox-os.org/redox-os/redoxfs/-/blob/master/src/mount/fuse.rs#L39)) in order to build an image with MacOS. We are upgrading from the fuse library and would like to use this option.

### DIFF
--- a/src/mnt/mount_options.rs
+++ b/src/mnt/mount_options.rs
@@ -28,6 +28,9 @@ pub enum MountOption {
     /// Enable permission checking in the kernel
     DefaultPermissions,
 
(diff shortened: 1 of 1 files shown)


cberner:
Hmm. I'm not sure about adding this, as it appears to be a MacFUSE only feature. Have you tried using `MountOption::CUSTOM("defer_permissions")`?

freewilll:
Ahh nice, I didn't realize `MountOption` was an option (pun not intended). I'll use that instead, thanks.

PR #235, opened 2023-03-17
Avoid copying reply buffer in `ReplyData`

jamesbornholt:
The reply already doesn't take ownership of the buffer, and it just ends up being written synchronously to the FUSE device, so there's no need to make a copy of it. Instead, the `Response` can just hold onto the slice and use it directly when building the IoVec for writing the response.

This is an alternative to https://github.com/cberner/fuser/pull/232 that is hopefully a little simpler while solving the same problem.

Fixes #232


cberner:
Merged. Thanks!

PR #236, opened 2023-04-24
Fix some compiler warnings

jamesbornholt:
* `#[must_use]` doesn't do anything on impls, only methods
* We hold onto a couple of arguments that aren't used for convenience


cberner:
Merged. Thanks!

PR #240, opened 2023-05-19
Fix clippy lints

rydrman:
Simply running `cargo clippy --fix` with rust `1.68.2` to fixup current lints that clippy is reporting.


rydrman:
Apologies for the noise, it seems the initial changeset was a bit aggressive

rydrman:
This appears to be a deeper rabbit hole than I anticipated, there are some good fixes here if you want to look closer, but I don't think I have the time to keep resolving these builds

Issue #216, opened 2022-08-02
Is there a good reason to `parse_options_from_args` as  a `pub(crate)` and not as a `pub` ?


ahmed-masud:
https://github.com/cberner/fuser/blob/3fb6586aaf2e4ef3162c59ae798728a1dcf41eee/src/mnt/mount_options.rs#L161

Would be nice to have this available outside :-) as it builds out `MountOption` reasonably nicely.

cberner:
Yes, that is only there for backwards compatibility with old versions of this crate. I plan to remove it in the future, and so don't want to make that format public.

PR #241, opened 2023-05-19
Make `MountOption` to/from `str` conversion public

### LINKED ISSUE #229: Export MountOption::from_str and option_to_string from mount_options.rs
Right now I need to copy/paste this function into my [code](https://github.com/anuvu/puzzlefs/blob/master/reader/src/lib.rs#L19), it would be useful if these two function were exported by the library. The use case is that I want to allow mount options from the command line.

rydrman:
The goal here is to make the to/from string conversion of the `MountOption` type publically accessible. In the process, I've also implemented standard traits rather than using custom functions where appropriate. 

I considered using `std::str::FromStr` instead of `From`, but that trait has an associated error type and the current function is infallible.

Closes #229

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1005,7 +1005,7 @@ pub fn spawn_mount<'a, FS: Filesystem + Send + 'static + 'a, P: AsRef<Path>>(
 ) -> io::Result<BackgroundSession> {
     let options: Option<Vec<_>> = options
         .iter()
-        .map(|x| Some(MountOption::from_str(x.to_str()?)))
+        .map(|x| Some(MountOption::from(x.to_str()?)))
         .collect();
     let options = options.ok_or(ErrorKind::InvalidData)?;
     Session::new(filesystem, mountpoint.as_ref(), options.as_ref()).and_then(|se| se.spawn())
--- a/src/mnt/mount_options.rs
+++ b/src/mnt/mount_options.rs
@@ -59,8 +59,8 @@ pub enum MountOption {
     to libfuse, and not part of the kernel ABI */
 }
 
-impl MountOption {
-    pub(crate) fn from_str(s: &str) -> MountOption {
+impl<'a> From<&'a str> for MountOption {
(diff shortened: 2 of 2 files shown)


cberner:
Thanks, but I don't think this is something I want to support. There's some context in this issue: https://github.com/cberner/fuser/issues/216

Instead, I encourage people to copy that argument parsing code into their application, as I think that's the best location for code that handles command line options.

rydrman:
Is there a reason that you don't want to support this? Are you open to any discussion on the matter?

The options themselves are defined as strings in the kernel/fuse docs and change quite infrequently. The implementation is also extremely simple and saves all of us from having to repeat the logic in multiple places.

cberner:
I don't think that argument parsing belongs in this crate. Decisions about argument parsing should be made by the downstream application, and there are often conventions to follow for their platform -- personally I think those names are poor and should be full words separated with hypens, but I don't want to force that on all my users.

Yes, I'm always open to discussion, but it probably won't change my mind. I'm a pretty firm believer in separation of concern :)

rydrman:
In my case, it's more serialization than argument parsing, since I am needing to construct a mount command from a set of options (yes, still interfacing with the CLI, I get that).

I suppose for me, it seems odd not to maintain a conversion for items that have a very clear and established string representation - put another way: by defining this enum, you are requiring that most of everyone using this library needs to convert from the existing standard names to this unique type. I appreciate the separation of concerns with command line parsing, but I don't think that's what you are supporting here. Rather, you are supporting conversion from/to the standard form of the values that you are trying to represent. This feels appropriate to me if you are going to define your own custom type for them.

That's all I'll say on the matter - if you still disagree then you can close this PR.

cberner:
yep, I see your point, but it goes against my design philosophy. I think the appropriate place for functionality like that would be a `fuse-utils` crate or the like. If you want to copy the parsing code and release it in a utils crate, I have no objections!

PR #242, opened 2023-06-05
Add examples cargo project files

tasleson:
I wanted to build some of the examples, to try them out and found it a bit tedious to get the cargo dependencies correct.

There is likely a way to use [workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html), to build these automatically when the main package gets built, but I didn't look into that as I'm not sure others would want it.


tasleson:
> Sorry, but I don't want to change this packaging. Feel free to submit the other cleanups though, if you want

No worries, I didn't realize you could build something from the examples directory with:
`cargo build --example simple --no-default-features` which I discovered in the CI output.

clean-ups: https://github.com/cberner/fuser/pull/243

Issue #246, opened 2023-06-15
`users` is unmaintained


oxalica:
https://rustsec.org/advisories/RUSTSEC-2023-0040

Seems we only use its `get{u,g}id` for mount flags, which can be replaced by [`rustix::process::get{u,g}id`](https://docs.rs/rustix/0.37.20/rustix/process/fn.getuid.html) or the [raw `libc::get{u,g}id`](https://docs.rs/libc/0.2.146/libc/fn.getuid.html).
https://github.com/cberner/fuser/blob/39d4177e809c7ee3b6757136fee8a28d5f41f040/src/mnt/fuse_pure.rs#L346-L347

PR #247, opened 2023-06-16
removing users dependency

TornaxO7:
This should fix https://github.com/cberner/fuser/issues/246.

**Note:**
I'm just a complete newby at this project and I thought that this could be quickly fixed, so please carefully look at the PR.


TornaxO7:
The error:

> error: package `log v0.4.19` cannot be built because it requires rustc 1.60.0 or newer, while the currently active rustc version is 1.57.0

do you have any suggestions how to fix that?

cberner:
Oh, just edit the *.Dockerfile to use a more recent version

TornaxO7:
done!

PR #258, opened 2023-08-24
Add cargo.toml to examples

noamraph:
I wanted to try the examples, and I wasn't sure how to do it. So I moved them to their own directories with `Cargo.toml` files, so now I can just run `cargo run` and they get compiled and run.

Is this a good idea?

Thanks!
Noam


wfraser:
You can already run the examples like `cargo run --example simple`. They also get built with the dependencies in `[dev-dependencies]` in the main Cargo.toml; it's not necessary to have a separate Cargo.toml for each one.

PR #275, opened 2024-03-10
Make Session::new consistent with mount2

daniel-levin:
Hiya @cberner!

I am a huge fan of this crate. It works very well for me. This PR is a small quality of life improvement that makes the `Session::new` method consistent with `mount2`/`spawn_mount2`. That is, you can now pass in anything that is `AsRef<Path>` as the mountpoint. Since `&Path` is `AsRef<Path>`, ~~this won't break existing callsites.~~ - clearly it will break some callsites where the compiler infers which `AsRef<T>` is the appropriate type for `String`.

### DIFF
--- a/examples/notify_inval_entry.rs
+++ b/examples/notify_inval_entry.rs
@@ -162,7 +162,7 @@ fn main() {
         timeout: Duration::from_secs_f32(opts.timeout),
     };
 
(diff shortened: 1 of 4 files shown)


daniel-levin:
Fixed failures and rebased.

PR #279, opened 2024-03-25
Add file handle for get_attr requests.

IamTheCarl:
This PR adds the file handle parameter to the `getattr` request.
This feature is necessary when implementing filesystem APIs that generate file contents on request, like `/proc`.
The correct length of the file must be returned by `getattr` for that specific instance of the generated file, otherwise most programs will not read the file correctly. A file handle is needed to know which instance we are getting attributes for.

The parameter is available in ABI 7-9 and above, and has been feature gated as such. When the feature is not enabled, the `getattr` method on Filesystems will always be passed None for the file descriptor.

### DIFF
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -58,7 +58,7 @@ impl Filesystem for HelloFS {
         }
     }
 
-    fn getattr(&mut self, _req: &Request, ino: u64, reply: ReplyAttr) {
+    fn getattr(&mut self, _req: &Request, ino: u64, _fd: Option<u64>, reply: ReplyAttr) {
         match ino {
             1 => reply.attr(&TTL, &HELLO_DIR_ATTR),
             2 => reply.attr(&TTL, &HELLO_TXT_ATTR),
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -58,7 +58,7 @@ impl Filesystem for HelloFS {
         }
     }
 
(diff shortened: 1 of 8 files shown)


examples/hello.rs
```diff
@@ -58,7 +58,7 @@ impl Filesystem for HelloFS {
         }
     }
 
-    fn getattr(&mut self, _req: &Request, ino: u64, reply: ReplyAttr) {
+    fn getattr(&mut self, _req: &Request, ino: u64, _fd: Option<u64>, reply: ReplyAttr) {
```
cberner, line 61: Can you name these `_fh` to match the rest of the code? Same comment for the other examples

cberner (changes requested):
One minor style comment

src/lib.rs
```diff
@@ -331,8 +331,11 @@ pub trait Filesystem {
     }
 
     /// Get file attributes.
-    fn getattr(&mut self, _req: &Request<'_>, ino: u64, reply: ReplyAttr) {
-        warn!("[Not Implemented] getattr(ino: {:#x?})", ino);
+    fn getattr(&mut self, _req: &Request<'_>, ino: u64, fd: Option<u64>, reply: ReplyAttr) {
```
cberner, line 334: You missed one. Otherwise this lgtm

cberner (changes requested):

cberner (approved):

IamTheCarl:
Thank you!

PR #282, opened 2024-04-10
chore(examples): fix HelloFS getattr signature

AleksanderGondek:
This commit adjusts the signature of HelloFS 'getattr' function, so that it is aligned with the Filesystem trait, thus making the example build.

In addition, the following minor miscellaneous changes were made to examples/hello.rs:
* dependency on libc crate was removed, to simplify the example and its usage
* unnecessary `env_logger::init()` call was removed


AleksanderGondek:
Nevermind, I should have RTFM.

PR #286, opened 2024-06-03
Add CI to run tests on macOS

FirelightFlagboy:
- Enable test `unmount_no_send` for macOS but skip it on CI

  This is due to the macFUSE's kernel extension that is not installed causing the test to not finish.

- Disable test `mount_unmount`, it seems macOS need an handshake of some sort for the mountpoint to be acknowledged.

- Fix unused variable when targeting macos.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,3 +63,23 @@ jobs:
 
       - name: Run tests
         run: INTERACTIVE="" make ${{ matrix.test_group }}
+
+  test-mac:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+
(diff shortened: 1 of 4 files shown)


cberner:
Hmm, since the MacFUSE extension doesn't work. Does this add meaningful coverage? It looks like it only runs the unit tests for the request parsing and such, which are already covered by the Linux CI

FirelightFlagboy:
> Does this add meaningful coverage?

If the meaningful tests for this crate are the mountpoint tests, since we can't run them on the macos runner then no.

But at least it check if we can compile to macos :shrug:

cberner:
Ya, unfortunately the real test coverage comes from the pjdfs and xfstests

PR #288, opened 2024-07-08
Add ValueParserFactory for MountOption

dr-bonez:
gated under the "clap" feature


cberner:
Hmm, I generally don't think fuser should dictate the format of command line arguments. Is there broad precedent for doing that in Rust crates?

dr-bonez:
In order to support `mount -t <fstype>` it needs to be able to parse mount options in that specific format. In this case, fuser isn't opinionated about anything, mount is.

cberner:
Ok ya, that's a good argument. Can you add a few tests?

PR #290, opened 2024-07-12
refactor(ll): adapt the `fuse-abi` crate

mkroening:
Hi there,

I have published the [`fuse-abi`](https://github.com/rust-osdev/fuse-abi) crate for autogenerated, up-to-date bindings to FUSE devices, which is also usable in `no_std` environments such as OS kernels.

This initial unclean port of `fuser` to `fuse-abi` seems to work in practice, according to CI.
The most important difference to `fuser::ll::fuse_abi` is that the crate's struct's fields are not feature-gated. This would require `fuser` to adapt another approach to access to versioned fields which change the struct's size. This does not happen a lot, but there are cases such as with `fuse_mknod_in`, which libfuse handles like [this](https://github.com/libfuse/libfuse/blob/fuse-3.16.2/lib/fuse_lowlevel.c#L1232-L1235).

If you are open to such an approach or have another idea to this issue, I'd be happy to continue working on this. I just wanted to test the waters before putting in more work. :)


cberner:
I prefer to minimize the number of dependencies in my crates, so don't want to merge this. Looks neat though, thanks for submitting!

PR #293, opened 2024-08-20
WIP: Add Multithreading Support

volfco:
Add Multithreading Support to fuser.

This implementation modifies `BackgroundSession` to accept a thread quantity. If this number is two or more, then n - 1 background workers will be created. There will always be a primary worker created, hence n - 1. 

`Session::worker` is called to clone the Session, which in turn calls `Channel::worker`. This method accepts the Mount object, and using `.session_fd` to return the Session FD, `fuse_fd_clone` is invoked to clone the file descriptor and run the needed ioctl call to setup the session. 

I can't take all the credit, as the main piece of code in channel.rs is from the Datenlord project where they are doing a native async_fuse library.

The code in question is from here: https://github.com/datenlord/datenlord/blob/d90fd43732373451207e56e9b9cd7eef9e7b53e1/src/async_fuse/fuse/session.rs#L73

Additional References:
* https://john-millikin.com/the-fuse-protocol#multi-threading
* https://fuse-devel.narkive.com/CodQceXu/how-could-fuse-support-parallel-opreation

TODO

- [ ] Clean up feature gate
- [ ] Add more function documentation 
- [ ] Add changelog entry
- [ ] Bump version

### DIFF
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -32,6 +32,7 @@ use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use std::{env, fs, io};
+use std::sync::Arc;
 
 const BLOCK_SIZE: u64 = 512;
 const MAX_NAME_LENGTH: u32 = 255;
@@ -243,9 +244,10 @@ impl From<InodeAttributes> for fuser::FileAttr {
 
 // Stores inode metadata data in "$data_dir/inodes" and file contents in "$data_dir/contents"
 // Directory data is stored in the file's contents, as a serialized DirectoryDescriptor
+#[derive(Clone)]
 struct SimpleFS {
     data_dir: String,
-    next_file_handle: AtomicU64,
+    next_file_handle: Arc<AtomicU64>,
     direct_io: bool,
     suid_support: bool,
 }
@@ -260,7 +262,7 @@ impl SimpleFS {
         {
             SimpleFS {
                 data_dir,
-                next_file_handle: AtomicU64::new(1),
+                next_file_handle: Arc::new(AtomicU64::new(1)),
                 direct_io,
                 suid_support,
             }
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -291,7 +291,7 @@ impl KernelConfig {
 /// implementations are provided here to get a mountable filesystem that does
 /// nothing.
 #[allow(clippy::too_many_arguments)]
-pub trait Filesystem {
+pub trait Filesystem: Clone {
```
colinmarc, line 294: Sorry for the random drive-by comment. In my opinion, it would be better to require `Send + Sync` instead of clone, and change the trait methods to take non-mutable `self`. Then it's up to the user how they handle mutability, if they need it. Cloning + mutability is a massive footgun.

examples/simple.rs
```diff
@@ -243,9 +244,10 @@ impl From<InodeAttributes> for fuser::FileAttr {
 
 // Stores inode metadata data in "$data_dir/inodes" and file contents in "$data_dir/contents"
 // Directory data is stored in the file's contents, as a serialized DirectoryDescriptor
+#[derive(Clone)]
 struct SimpleFS {
     data_dir: String,
-    next_file_handle: AtomicU64,
+    next_file_handle: Arc<AtomicU64>,
```
colinmarc, line 250: For example, `Arc<AtomicU64>` is a totally reasonable way to handle mutability, here - but this example could have easily used a normal `u64` before these changes and still derived `Clone`, leading to coinciding u64s.

src/lib.rs
volfco: Thanks for the feedback!

I need to figure out exactly what you're referring to and understand it, but that does make sense and does seem better. Something fun to do this weekend!

theunkn0wn1: if a type is `Send`, it can be safely moved between threads.

if a type is `Sync`, it can be safely shared between threads.
see https://doc.rust-lang.org/nomicon/send-and-sync.html?highlight=Send#send-and-sync

volfco:
> Do mind explaining why you picked the multithreading programming model rather than async/await programming model?

Because that would be a major refactor of the code base, and to do it properly everything would need to be rewritten

cberner:
Closing, since this hasn't been worked on in a while

PR #294, opened 2024-09-04
Add an ioctl example

colinmarc:
This matches libfuse's example:

https://github.com/libfuse/libfuse/blob/d30247c36dadd386b994cd47ad84351ae68cc94c/example/ioctl.c

Unfortunately, it doesn't work at present. After mounting at `/tmp/foobar`, you can test it by compiling libfuse's ioctl client:

```
git clone https://github.com/libfuse/libfuse
cd libfuse/example
gcc -Wall ioctl_client.c `pkg-config fuse3 --cflags --libs` -o ioctl_client
./ioctl_client /tmp/foobar/fioc
```
This example returns `ENOTTY` without ever hitting the ioctl handler:

```
openat(AT_FDCWD, "/tmp/foobar/fioc", O_RDWR) = 3
ioctl(3, _IOC(_IOC_READ, 0x45, 0, 0x8), 0x7fffaf4404b0) = -1 ENOTTY (Inappropriate ioctl for device)
```

Whereas the libfuse example works:

```
ioctl(3, _IOC(_IOC_READ, 0x45, 0, 0x8), 0x7ffdb6c5b970) = 0
```

I'm happy to do the work for a fix, but I haven't been able to figure out where the bug is yet :)

### DIFF
--- a/examples/ioctl.rs
+++ b/examples/ioctl.rs
@@ -1,0 +1,209 @@
+// This example requires fuse 7.11 or later. Run with:
+//
+//   cargo run --example ioctl --features abi-7-11 /tmp/foobar
+
+use clap::{crate_version, Arg, ArgAction, Command};
+use fuser::{
+    FileAttr, FileType, Filesystem, MountOption, ReplyAttr, ReplyData, ReplyDirectory, ReplyEntry,
+    Request,
+};
+use libc::{EINVAL, ENOENT};
+use log::debug;
+use std::ffi::OsStr;
+use std::time::{Duration, UNIX_EPOCH};
+
+const TTL: Duration = Duration::from_secs(1); // 1 second
+
+struct FiocFS {
+    content: Vec<u8>,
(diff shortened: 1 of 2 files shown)


colinmarc:
Ok, figured it out. I needed `--features abi-7-11` and to generate the ioctl codes correctly.

examples/ioctl.rs
```diff
@@ -0,0 +1,209 @@
+// This example requires fuse 7.11 or later. Run with:
+//
+//   cargo run --example ioctl --features abi-7-11 /tmp/foobar
```
cberner, line 3: In the Cargo.toml file, you should add this as a `required-features` like some of the other examples

cberner (changes requested):

examples/ioctl.rs
colinmarc: Done.

cberner (approved):

Issue #300, opened 2024-10-06
Extract `Channel` FD and self-mount


colinmarc:
Hi, I have a use case (containers) where I need to get the session FD and do the mounting myself. I took a look at changing the API to support this, but unfortunately, it seems like `mnt::Mount` and `Session` are pretty tied together at the moment.

Before I propose any significant API changes, do you have an idea how I could achieve this?

Thanks!

colinmarc:
I poked around a bit more. First, let me clarify roughly what I'd like to do:

1. Create a `Session` with my `FileSystem`.
2. Call [`AsFd::as_fd`](https://doc.rust-lang.org/stable/std/os/fd/trait.AsFd.html) on the `Session`. Then, take that FD, and pass it into my own mount(2) call with `fd=123`.
3. Call `Session::run` to proceed as normal.

Fundamentally, this seems like it would require splitting `Session` into two types: one which holds the `/dev/fuse` FD (the `Channel`), and one which knows the mountpoint and calls `umount` when it's dropped.

However, The fact that a session is also a mountpoint, as far as the API is concerned, seems to stem from `libfuse`. That's because `fuse_session_new` takes the mount args already. So while the above change would be fairly trivial with `fuse_pure`, it's not easy with `libfuse`.

I don't suppose you'd consider dropping support for `libfuse`? :sweat_smile:

colinmarc:
Huh, there's also a (very wacky) libfuse3 way to do this: https://github.com/libfuse/libfuse/pull/291

cberner:
Can you tell me more about the use case? I'm trying to understand if this is common enough that it should be support in fuser

colinmarc:
> Can you tell me more about the use case? I'm trying to understand if this is common enough that it should be support in fuser

FUSE plays a special role in containers, where it can be used to simulate [kernel-level stuff like sysfs](https://github.com/lxc/lxcfs) or [character devices](https://crates.io/crates/southpaw). This is [specifically permitted](https://github.com/torvalds/linux/commit/4ad769f3c346ec3d458e255548dec26ca5284cf6) for unprivileged container runtimes by the kernel. However, various security safeguards mean that it requires a very specific dance:

 1. `/dev/fuse` has to be opened inside the container (inside the user namespace where the mount call happens)
 2. The `mount` (or `fsconfig`/`move_mount`) syscalls have to be done inside the container as well
[...]

PR #303, opened 2024-10-14
Allow users to create Notifier objects when handling requests

colinmarc:
This is often useful in `poll` implementations. It also comes into play in #300, because your initialization and setup (when you can call `Session::notifier`) is often separate from creating the session.

### DIFF
--- a/examples/poll.rs
+++ b/examples/poll.rs
@@ -23,7 +23,7 @@ use libc::{EACCES, EBADF, EBUSY, EINVAL, ENOENT, ENOTDIR};
 
 use fuser::{
     consts::{FOPEN_DIRECT_IO, FOPEN_NONSEEKABLE, FUSE_POLL_SCHEDULE_NOTIFY},
-    FileAttr, FileType, MountOption, Request, FUSE_ROOT_ID,
+    FileAttr, FileType, MountOption, PollHandle, Request, FUSE_ROOT_ID,
 };
 
 const NUMFILES: u8 = 16;
@@ -251,7 +251,7 @@ impl fuser::Filesystem for FSelFS {
         _req: &Request,
         _ino: u64,
         fh: u64,
-        kh: u64,
+        ph: PollHandle,
         _events: u32,
         flags: u32,
         reply: fuser::ReplyPoll,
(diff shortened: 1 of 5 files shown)


cberner:
This doesn't seem like the right place to create a `Notifier` to me. Do you have an example of how you're using it?

colinmarc:
Sure, for example here: https://github.com/colinmarc/southpaw/blob/main/src/fs.rs#L350-L354

Each `open` (which results in an async read) gets its own notifier, so I don't have to pass them around everywhere.

cberner:
I haven't actually used the notification interface myself, but my understanding is that a `Notifier` is a communication channel that can push messages to the kernel for updates to various inodes, dir entries...etc. Because it's not specific to the inode being opened in `open()` adding it to the `Request` seems off to me.

If it was available in `Filesystem::init()` would that solve your case? Looking at the code again, I'm not sure that `Session::notifier()` is the right place to add it.

colinmarc:
> If it was available in Filesystem::init() would that solve your case? Looking at the code again, I'm not sure that Session::notifier() is the right place to add it.

I guess so, but I'd just be wrapping it in an `Arc` and passing it around some more. IMO `Notifiers` should be cheap, cloneable (they aren't `Clone` right now) and readily available. And I think the semantics should match FUSE's as closely as possible.

If you'd like to provide a tighter interface, instead, we could create a scoped object, something like

```rust
struct PollHandle {
    pub ph: u64
    notifier: Notifier
}

impl PollHandle {
    pub fn notify(self) -> io::Result<()> {
        todo!()
    }
}
```

And pass it in instead of `ph: u64` to the `poll()` call.

cberner:
That does seem better to me. I haven't used notification myself, but as I understand it there are two distinct use cases:
1) inode and directory cache management. For this case the current interface seems more or less right to me, because you need a `Notifier` client that you can hold on to and push notifications to the kernel independent of any requests coming in. Although passing it into `init()` might be better than retrieving it from `Session`
2) responding to `poll()` requests. In this case I agree the current interface isn't ideal. I like your suggestion of passing in a `PollHandle` that can be notified.

Are there any other cases I'm missing?

The thing I want to avoid is adding a notifier to all requests, because many requests don't need a notifier

colinmarc:
Updated to add the `PollHandle` type.

cberner (approved):
lgtm!

WilliamVenner:
Hi, I tried using the `Notifier` from within a FUSE handler (specifically for an `unlink` request) - trying to send a notification  (`inval_entry`, `inval_inode`, `delete`, etc.) causes a deadlock. I wanted to investigate why but unfortunately gdb hangs infinitely as well so I decided to move on. How is the Notifier intended to be used for that use case? My workaround for now is to just have a background thread that dispatches queued up notifications, but it would be nice if I could do that in sync with my actual filesystem ops.

PR #304, opened 2024-10-14
Allow wrapping an existing /dev/fuse file descriptor

### LINKED ISSUE #300: Extract `Channel` FD and self-mount

colinmarc:
This is important for container runtimes, which need to do a special namespace mount dance.

Fixes #300. This is a less invasive version of #301, but I like that one better.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -9,6 +9,7 @@ use libc::{EAGAIN, EINTR, ENODEV, ENOENT};
 use log::{info, warn};
 use nix::unistd::geteuid;
 use std::fmt;
+use std::os::fd::{AsFd, BorrowedFd, OwnedFd};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 use std::thread::{self, JoinHandle};
@@ -31,10 +32,15 @@ pub const MAX_WRITE_SIZE: usize = 16 * 1024 * 1024;
 /// up to MAX_WRITE_SIZE bytes in a write request, we use that value plus some extra space.
 const BUFFER_SIZE: usize = MAX_WRITE_SIZE + 4096;
 
-#[derive(Debug, Eq, PartialEq)]
-pub(crate) enum SessionACL {
+#[derive(Default, Debug, Eq, PartialEq)]
+/// How requests should be filtered based on the calling UID.
+pub enum SessionACL {
+    /// Allow requests from any user. Corresponds to the `allow_other` mount option.
     All,
+    /// Allow requests from root. Corresponds to the `allow_root` mount option.
     RootAndOwner,
+    /// Allow requests from the owning UID. This is FUSE's default mode of operation.
+    #[default]
     Owner,
 }
 
@@ -46,9 +52,7 @@ pub struct Session<FS: Filesystem> {
     /// Communication channel to the kernel driver
     pub(crate) ch: Channel,
     /// Handle to the mount.  Dropping this unmounts.
-    mount: Arc<Mutex<Option<Mount>>>,
-    /// Mount point
-    mountpoint: PathBuf,
+    mount: Arc<Mutex<Option<(PathBuf, Mount)>>>,
     /// Whether to restrict access to owner, root + owner, or unrestricted
     /// Used to implement allow_root and auto_unmount
     pub(crate) allowed: SessionACL,
@@ -62,6 +66,12 @@ pub struct Session<FS: Filesystem> {
     pub(crate) initialized: bool,
     /// True if the filesystem was destroyed (destroy operation done)
     pub(crate) destroyed: bool,
+}
+
+impl<FS: Filesystem> AsFd for Session<FS> {
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        self.ch.as_fd()
+    }
 }
 
 impl<FS: Filesystem> Session<FS> {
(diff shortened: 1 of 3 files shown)


src/session.rs
```diff
@@ -111,9 +120,21 @@ impl<FS: Filesystem> Session<FS> {
         })
     }
 
-    /// Return path of the mounted filesystem
-    pub fn mountpoint(&self) -> &Path {
-        &self.mountpoint
+    /// Wrap an existing /dev/fuse file descriptor. This doesn't mount the
+    /// filesystem anywhere; that must be done separately.
+    pub fn from_fd(filesystem: FS, fd: OwnedFd, acl: SessionACL) -> Self {
```
cberner, line 125: Instead of `SessionACL` I think this should take `&[MountOption]` and in the documentation state that it should be the same set of options that was used to create the mount and FD.

Can you also add to the documentation that the caller is responsible for unmounting the mountpoint?

src/session.rs
```diff
@@ -64,6 +68,12 @@ pub struct Session<FS: Filesystem> {
     pub(crate) destroyed: bool,
 }
 
+impl<FS: Filesystem> AsFd for Session<FS> {
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        self.ch.as_fd()
+    }
+}
+
```
cberner, line 76: I think it would be better to have a `mount_to_fd()` function similar to `mount()` that instead returns the FD. That better separates the concerns of creating the mountpoint, which won't even need to have a `Filesystem` implementation.

src/session.rs
```diff
@@ -31,10 +32,15 @@ pub const MAX_WRITE_SIZE: usize = 16 * 1024 * 1024;
 /// up to MAX_WRITE_SIZE bytes in a write request, we use that value plus some extra space.
 const BUFFER_SIZE: usize = MAX_WRITE_SIZE + 4096;
 
-#[derive(Debug, Eq, PartialEq)]
-pub(crate) enum SessionACL {
+#[derive(Default, Debug, Eq, PartialEq)]
+/// How requests should be filtered based on the calling UID.
+pub enum SessionACL {
```
cberner, line 37: I'd rather not make this public. See my comment about `&[MountOption]`. Does that seem like it will work?

cberner (changes requested):

src/session.rs
colinmarc: This is just for symmetry - I'm not using it, so I can remove it if you don't like it. Note that the stdlib equivalents all support this (File, various sockets, etc).

src/session.rs
colinmarc: > Instead of `SessionACL` I think this should take `&[MountOption]` and in the documentation state that it should be the same set of options that was used to create the mount and FD.

Hm, I don't agree. A mount option like `fsname` has no effect here and will be ignored.

What if I added `SessionACL::from_mount_options(options: impl AsRef<[MountOption]>) -> Self`?

src/session.rs
```diff
@@ -47,8 +53,6 @@ pub struct Session<FS: Filesystem> {
     ch: Channel,
     /// Handle to the mount.  Dropping this unmounts.
     mount: Arc<Mutex<Option<Mount>>>,
-    /// Mount point
-    mountpoint: PathBuf,
```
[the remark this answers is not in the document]
colinmarc: What is the utility of getting the mountpoint back out, after you just passed it in? Especially if it returns an `Option<PathBuf>`?

src/session.rs
cberner: Ok ya, I take it back. I'd been thinking there might be other `MountOption`s that are useful, but looking through the code again I see that they're all consumed when the mount is created.

src/session.rs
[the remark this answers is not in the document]
cberner: The getter doesn't seem useful, but it seems nice to have the logging

src/session.rs
cberner: How would someone use it though?

The reason it doesn't seem right to me is that a `Session` represents a mounted filesystem, borrowing its fd, and reading/writing it would interfere with the `Filesystem` that is mounted and feels like it breaks the encapsulation.

colinmarc: > How would someone use it though?

By passing it into `mount(2)`/`fsconfig(2)` to mount the session somewhere.

In my use case, you need to mount first and then create the session. But it's not hard to imagine someone needing to do it the other way around.

colinmarc: Zooming out a bit:

I would like to submit that FUSE is a very low-level concept, and it's often used as a hack-on-demand. So people using it probably know what they're doing, and may have complicated requirements. A fuse session is a `/dev/fd` fuse - therefore it fulfills the requirements of `AsFd`.

I do see the concerns around breaking the encapsulation, but `Channel` is not actually stateful, right? So it's really very similar to `UnixStream` or whatever. If you call `AsFd` on a `UnixStream` and drop down to `nix::io::write` or whatever, you probably have a very specific use case that the stdlib authors didn't think of.

src/session.rs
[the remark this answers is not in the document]
colinmarc: Added the logging back.

src/session.rs
cberner: Good points. Thanks for explaining the use case!

cberner (approved):

PR #307, opened 2024-10-21
Fix `macFuse` `fuse_rename_in` missing fields

FirelightFlagboy:
`MacFuse@4.8` define the struct `fuse_rename_in` like so:

```c
struct fuse_rename_in {
	__u64	newdir;
#ifdef __APPLE__
	__u32	flags;
	__u32	padding;
#endif
};
```

> [`fuse_rename_in`](https://github.com/osxfuse/fuse/blob/6f7322893456f6ff9db145f096b9bfc2ba95d627/include/fuse_kernel.h#L446-L452)

But on `fuser` side, we only have the field `newdir`:

```rust
pub struct fuse_rename_in {
    pub newdir: u64,
}
```

> [`fuse_rename_in`](https://github.com/cberner/fuser/blob/v0.14.0/src/ll/fuse_abi.rs#L584-L588)

This result in `name` and `newname` (of `Rename`) being at best empty Path or invalid valid depending on the value of the field `flags` & `padding`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ pkg-config = { version = "0.3.14", optional = true }
 default = ["libfuse"]
 libfuse = ["pkg-config"]
 serializable = ["serde"]
+macfuse-4-compat = []
 abi-7-9 = []
 abi-7-10 = ["abi-7-9"]
 abi-7-11 = ["abi-7-10"]
--- a/build.rs
+++ b/build.rs
@@ -13,6 +13,7 @@ fn main() {
                 .is_ok()
             {
                 println!("cargo:rustc-cfg=feature=\"libfuse2\"");
+                println!("cargo:rustc-cfg=feature=\"macfuse-4-compat\"");
             } else {
                 pkg_config::Config::new()
                     .atleast_version("2.6.0")
--- a/src/ll/fuse_abi.rs
+++ b/src/ll/fuse_abi.rs
@@ -586,9 +586,9 @@ pub struct fuse_mkdir_in {
 pub struct fuse_rename_in {
     pub newdir: u64,
     #[cfg(target_os = "macos")]
-    pub _flags: u32,
+    pub flags: u32,
     #[cfg(target_os = "macos")]
-    pub _padding: u32,
+    pub padding: u32,
 }
 
 #[repr(C)]
--- a/src/ll/fuse_abi.rs
+++ b/src/ll/fuse_abi.rs
@@ -585,9 +585,9 @@ pub struct fuse_mkdir_in {
 #[derive(Debug, FromBytes, FromZeroes)]
 pub struct fuse_rename_in {
     pub newdir: u64,
-    #[cfg(target_os = "macos")]
+    #[cfg(feature = "macfuse-4-compat")]
     pub flags: u32,
-    #[cfg(target_os = "macos")]
+    #[cfg(feature = "macfuse-4-compat")]
     pub padding: u32,
 }
 
--- a/src/ll/fuse_abi.rs
+++ b/src/ll/fuse_abi.rs
@@ -585,6 +585,10 @@ pub struct fuse_mkdir_in {
 #[derive(Debug, FromBytes, FromZeroes)]
 pub struct fuse_rename_in {
     pub newdir: u64,
+    #[cfg(target_os = "macos")]
+    pub _flags: u32,
+    #[cfg(target_os = "macos")]
+    pub _padding: u32,
 }
 
 #[repr(C)]


FirelightFlagboy:
<details>

<summary>Problem & Investigation</summary>

I've create a MemFS with `fuser` and I'm having issue with the rename operation on MacOS (using MacFuse@4.8) where the rename request is provided with empty filenames:

```shell
touch bar
mv bar zar
```

Result in the following Request log:

```txt
FUSE(  4) ino 0x00..01 RENAME src FilenameDir { dir: Inode(1), name: "" }, dest FilenameDir { dir: Inode(1), name: "" }
```

I've added a log that dump the content of `data` before it's decoded into the Rename Request:

```txt
Parsing opcode FUSE_RENAME, data: AQAAAAAAAAAAAAAAAAAAAGJhcgB6YXIA
```

By decoding the data, we get:

```shell
$ echo AQAAAAAAAAAAAAAAAAAAAGJhcgB6YXIA | base64 --decode | xxd
00000000: 0100 0000 0000 0000 0000 0000 0000 0000  ................
00000010: 6261 7200 7a61 7200                      bar.zar.
```

So we have the filenames that are provided, that mean the issue is during the parsing of the low-level request from macFuse.
`fuser` define the Rename request like so:

```rust
#[derive(Debug)]
pub struct Rename<'a> {
    header: &'a fuse_in_header,
    arg: &'a fuse_rename_in,
    name: &'a Path,
    newname: &'a Path,
}
```

> [`Rename`](https://github.com/cberner/fuser/blob/424bbcaa30a586a6dc99b5019f5e9b9f7755a31f/src/ll/request.rs#L581-L587)

```rust
#[repr(C)]
#[derive(Debug, FromBytes, FromZeroes)]
pub struct fuse_rename_in {
    pub newdir: u64,
}
```

> [`fuse_rename_in`](https://github.com/cberner/fuser/blob/v0.14.0/src/ll/fuse_abi.rs#L584-L588)

Parsing the provided data by MacFuse result in:

```rst
+---------------------+------------+---------------+------------------------------------+
| Rename                                                                                |
+=====================+============+===============+====================================+
| fuse_rename_in      |                                                                 |
+---------------------+------------+---------------+------------------------------------+
| newdir: u64         | name: Path | newname: Path | _unused_                           |
+---------------------+------------+---------------+------------------------------------+
| 0100 0000 0000 0000 | 00         | 00            | 0000 0000 0000 6261 7200 7a61 7200 |
+---------------------+------------+---------------+------------------------------------+
```

`fuse_rename_in` is defined as the following on MacFuse:

```c
struct fuse_rename_in {
	__u64	newdir;
#ifdef __APPLE__
	__u32	flags;
	__u32	padding;
#endif
};
```

> [`fuse_rename_in`](https://github.com/osxfuse/fuse/blob/6f7322893456f6ff9db145f096b9bfc2ba95d627/include/fuse_kernel.h#L446-L452)

If we map the data received, everything seems to _click into place_:

```rst
+---------------------+------------+--------------+------------+---------------+
| Rename                                                                       |
+=====================+============+==============+============+===============+
| fuse_rename_in                                  |                            |
+---------------------+------------+--------------+------------+---------------+
| newdir: u64         | flags: u32 | padding: u32 | name: Path | newname: Path |
+---------------------+------------+--------------+------------+---------------+
| 0100 0000 0000 0000 | 0000 0000  | 0000 0000    | 6261 7200  | 7a61 7200     |
+---------------------+------------+--------------+------------+---------------+
```

</details>

cberner:
Is this specific to version 4.8? It seems like this needs to be conditional on the macfuse version, otherwise older versions will break

FirelightFlagboy:
> Is this specific to version 4.8? It seems like this needs to be conditional on the macfuse version, otherwise older versions will break

I cannot say for sure, their git history is not linear. that was added by this commit: https://github.com/osxfuse/fuse/commit/f95e93946e0577ad91a055554773a5c0ff7e8f70 made in 2018

FirelightFlagboy:
After reading the release note, It seems that additional field where added in 4.0.0: https://github.com/osxfuse/osxfuse/releases/tag/macfuse-4.0.0

> Add support for renamex_np(2) on macOS 10.12 and later (VOL_CAP_INT_RENAME_SWAP and VOL_CAP_INT_RENAME_EXCL).
>
> Add new callbacks to libfuse
>
> ```c
> void renamex(fuse_req_t req, fuse_ino_t parent, const char *name, fuse_ino_t newparent, const char *newname, unsigned int flags);
> int renamex(const char *from, const char *to, unsigned int flags);
> ```
> 
> Add new callback to macFUSE.framework
>
> ```c
> typedef NS_OPTIONS(NSUInteger, GMUserFileSystemMoveOption) {
>    GMUserFileSystemMoveOptionSwap      = 1 << 0,
>    GMUserFileSystemMoveOptionExclusive = 1 << 1
> };
> ```

- (BOOL)moveItemAtPath:(NSString *)source toPath:(NSString *)destination options:(GMUserFileSystemMoveOption)options error:(NSError **)error;

cberner:
Cool, thanks!

cberner:
Done. Just published 0.15.0

PR #313, opened 2024-11-06
Replace features libfuse2,libfuse3 with a dedicated build cfg value

### LINKED ISSUE #311: Build warnings due to undeclared features "libfuse2", "libfuse3"
Hey! We currently maintain a forked version of `fuser` for our own use and are building on newer Rust versions (1.82).

From Rust v1.80, cfg values will be checked to ensure they match all known possible values. https://blog.rust-lang.org/2024/05/06/check-cfg.html

Fuser makes use of the features `libfuse2` and `libfuse3` when conditionally compiling code for each implementation of libfuse. This leads to a lot of warnings when building with newer versions of Rust.

I'd like to propose that we move this conditional logic to be a new cfg values (not feature) which we can declare so that Rust is aware of the possible values. Alternatively, the 'easy' option is to just declare `libfuse2` and `libfuse3` as features however this doesn't seem right, as these aren't meant to be customer-facing features.

I've created pull request https://github.com/cberner/fuser/pull/313 that implements this change, however I'm open to alternatives.

dannycjones:
From Rust v1.80, cfg values will be checked to ensure they match all known possible values. https://blog.rust-lang.org/2024/05/06/check-cfg.html

fuser makes use of the features libfuse2 and libfuse3 when conditionally compiling code for each implementation of libfuse. This leads to a lot of warnings when building fuser with newer versions of Rust (v1.80+).

This change moves this conditional logic to be a new dedicated cfg field (not features) which are declared so that Rust is aware of the possible values.

An alternative to this would be to just declare libfuse2 and libfuse3 as features however this doesn't seem right, as these aren't meant to be customer-facing features.

This change was originally proposed in #311.

Fixes #311

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,18 +1,25 @@
 fn main() {
+    // Register rustc cfg for switching between mount implementations.
+    // When fuser MSRV is updated to v1.77 or above, we should switch from 'cargo:' to 'cargo::' syntax.
+    println!("cargo:rustc-check-cfg=cfg(fuser_mount_impl, values(\"pure-rust\", \"libfuse2\", \"libfuse3\"))");
+
     #[cfg(all(not(feature = "libfuse"), not(target_os = "linux")))]
     unimplemented!("Building without libfuse is only supported on Linux");
 
+    #[cfg(not(feature = "libfuse"))]
+    {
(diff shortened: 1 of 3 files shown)


cberner:
Thanks!

PR #323, opened 2024-11-27
Fix ci cache concurrency issue

FirelightFlagboy:
Using the same cache key across multiple jobs can cause issues if we make assumption about it's content (like installing tools if not hit on the cache).

That cache was also forgetting caching `rustup` folder


cberner:
Did you intend to include the MD file? Looks like that's meant to be filed as an Issue?

FirelightFlagboy:
> Did you intend to include the MD file? Looks like that's meant to be filed as an Issue?

No :sweat_smile:

cberner:
Thanks!

PR #334, opened 2025-04-27
Fix several links in README.md

Mottl:
There is no issues with macOS, so probably we can remove "untested" term from README.md. I've also updated some links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,27 +63,22 @@ sudo yum install fuse-devel pkgconfig
 
 ### macOS (untested)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,29 +61,24 @@ sudo apt-get install libfuse-dev pkg-config
 sudo yum install fuse-devel pkgconfig
 ```
 
-### macOS (untested)
+### macOS
```
cberner, line 64: There's no CI for MacOS, so let's leave this as saying "untested"

cberner (changes requested):

cberner (approved):

PR #339, opened 2025-05-20
Fix macos flags to u64

daniel-white:
f67981 changed the flags for other oses to use `u64`. mac was neglected and failed to compile.


rust-toolchain
```diff
@@ -1 +1 @@
-1.74
+1.87.0
```
cberner, line 1: Let's revert this

cberner (changes requested):
See comment

PR #342, opened 2025-05-29
Remove flag and padding in rename operation when using libfuse

nicolaspapp:


cberner:
This seems like the wrong fix to me. Linux is the only supported OS with `not(libfuse)`, so this config can never be reached on MacOS. Should we just revert that flags field?

cberner (changes requested):
This seems like the wrong fix to me. Linux is the only supported OS with not(libfuse), so this config can never be reached on MacOS. Should we just revert that flags field?

FirelightFlagboy:
Maybe we just need to remove that line from `build.rs`: https://github.com/cberner/fuser/blob/a760aaeefee86d18bf136fc4de4c1d95cc600b29/build.rs#L23 ?

nicolaspapp:
I see. I was assuming that at some point in the future, supporting Mac without libfuse could be an option.

Regarding @FirelightFlagboy proposal, I think removing the fields would be the simplest solution. If Mac without libfuse is not an option, those fields are never going to be used. We can probably just get read off the fields and the `macfuse-4-compat` flag altogether.

Another option, which seems a lot more complicated but could be useful if you expect, at some point in time, having a pure Rust implementation for Mac, could be the following:
`#[cfg(target_os = "macos", not(feature = "libfuse"), feature = "macfuse-4-compat")]`

Since:
- It disables when libfuse is enabled since it does not use this flags.
- It will automatically support start using those flags if Mac without libfuse is ever an option
- You can use the feature tag `macfuse-4-compat` for macfuse < 4

@cberner your call. They are all fine for me, my only objective here is just to get the rename operation working on Mac.

FirelightFlagboy:
It still surprising, because macFuse still defined those field: https://github.com/osxfuse/fuse/blob/2f5d04ada43b42bde3c0ce49c5a5d977bce2b0f2/include/fuse_kernel.h#L446-L452

nicolaspapp:
Yeah, it is there on the kernel, but not on libfuse: https://github.com/libfuse/libfuse/blob/fc71597c8dba7ca8da474cbae3509552e30cbd46/include/fuse_kernel.h#L752

I haven't take a look at the libfuse implementation, my guess is that the library is truncating those 2 fields

cberner:
@nicolaspapp what MacFuse version are you using? It looks like the current code was to fix an issue with MacFuse >=4.0 (https://github.com/cberner/fuser/pull/307). However, now that I'm looking at build.rs I think the problem might be this line: https://github.com/cberner/fuser/blob/9f6ced73a36f1d99846e28be9c5e4903939ee9d5/build.rs#L17 we're enabling the `macfuse-4-compat` feature for MacFuse >= 2.6, which seems like it will break for 2.6 <= x < 4.0

nicolaspapp:
I am using the latest. Version 5.0.1.

I believe it might be related to to what happens on [fuse_lowlevel.c:1304](https://github.com/osxfuse/fuse/blob/master/lib/fuse_lowlevel.c#L1304)

image

`FUSE_COMPAT_RENAME_IN_SIZE` introduces an 8 bytes offset, which is the size of `flags` and `padding` variables

nicolaspapp:
@FirelightFlagboy under what situations was this initially breaking? Because what I have so far in the branch is basically reverting [#307](https://github.com/cberner/fuser/pull/307).

Based on what I see on the actual code `fuse_rename_in` should account for flags and padding only if `FUSE_CAP_RENAME_SWAP` is enabled. Now, it is not clear to me yet where that could be being set.

FirelightFlagboy:
> @FirelightFlagboy under what situations was this initially breaking? Because what I have so far in the branch is basically reverting [#307](https://github.com/cberner/fuser/pull/307).

The problem + investigation is detailed in the first comment: https://github.com/cberner/fuser/pull/307#issuecomment-2426103530

> Based on what I see on the actual code `fuse_rename_in` should account for flags and padding only if `FUSE_CAP_RENAME_SWAP` is enabled. Now, it is not clear to me yet where that could be being set.

Maybe `FUSE_CAP_RENAME_SWAP` was enabled by default before, and now it's not :thinking: 
I've tried previous version of macFuse that worked, and now they do not work so I suspect something have changed on the OS side

nicolaspapp:
@FirelightFlagboy I found something...

This is what I got so far:

1. Looks they first added experimental renamex support, and this is when they added the fields: https://github.com/macfuse/library/commit/f95e93946e0577ad91a055554773a5c0ff7e8f70

2. Then, later added the official support, adding that flag: https://github.com/macfuse/library/commit/3ec24b88be617ada78ac7f288836d7010bf984f8

I believe the first one was introduced at macFUSE 4.0.0. The second one seems to match 4.0.3 release description:


image



I think it might have been the default at some point, but ultimately what defines the FUSE_CAP_RENAME_SWAP is whether the `renamex` operation is defined in the FileSystem as shown [here](https://github.com/osxfuse/fuse/blob/master/lib/fuse_lowlevel.c#L2041)

I'm not sure why you were still seeing this issue in 4.8

PR #390, opened 2025-09-08
fix: Fix leaks and reads of unitialized memory in libfuse

In-line:
Found in https://github.com/ilya-zlobintsev/LACT/pull/740


cberner (approved):

cberner:
Thanks!

cberner:
Fine with me. I believe this only happens when the "libfuse" feature is enabled, so maybe recommend that people avoid that feature as a workaround

PR #392, opened 2025-09-10
Format, fix, document, Etc 02/15

rarensu:
[Clippy says](https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown) it's better to use backticks in doc strings. 
It formats prettier.

Previous PR in sequence was #391


cberner (approved):

cberner:
Merged. thanks!

PR #401, opened 2025-09-15
Format, fix, document, Etc 07/15

rarensu:
Improved documentation for many items. Update: removed unused Result type. Update: properly using the Result type, where appropriate, while avoiding some avoidable panics.

Previous PR in sequence was #400 
Next PR in sequence was #402 

An additional PR #405 branches from this point.

### DIFF
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -328,7 +328,7 @@ impl Reply for ReplyWrite {
 }
 
 impl ReplyWrite {
-    /// Reply to a request with the given open result
+    /// Reply to a request with the number of bytes written
     pub fn written(self, size: u32) {
         self.reply.send_ll(&ll::Response::new_write(size));
     }
@@ -356,7 +356,7 @@ impl Reply for ReplyStatfs {
 }
 
 impl ReplyStatfs {
-    /// Reply to a request with the given open result
+    /// Reply to a statfs request with filesystem information
     #[allow(clippy::too_many_arguments)]
     pub fn statfs(
         self,
@@ -397,7 +397,7 @@ impl Reply for ReplyCreate {
 }
 
 impl ReplyCreate {
-    /// Reply to a request with the given entry
+    /// Reply to a request with a newly created file entry and its newly open file handle
     /// # Panics
     /// When attempting to use kernel passthrough. Use `opened_passthrough()` instead.
     pub fn created(self, ttl: &Duration, attr: &FileAttr, generation: u64, fh: u64, flags: u32) {
@@ -436,7 +436,7 @@ impl Reply for ReplyLock {
 }
 
 impl ReplyLock {
-    /// Reply to a request with the given open result
+    /// Reply to a request with a file lock
     pub fn locked(self, start: u64, end: u64, typ: i32, pid: u32) {
         self.reply.send_ll(&ll::Response::new_lock(&ll::Lock {
             range: (start, end),
@@ -468,7 +468,7 @@ impl Reply for ReplyBmap {
 }
 
 impl ReplyBmap {
-    /// Reply to a request with the given open result
+    /// Reply to a request with a bmap
     pub fn bmap(self, block: u64) {
         self.reply.send_ll(&ll::Response::new_bmap(block));
     }
@@ -496,7 +496,7 @@ impl Reply for ReplyIoctl {
 }
 
 impl ReplyIoctl {
-    /// Reply to a request with the given open result
+    /// Reply to a request with an ioctl
     pub fn ioctl(self, result: i32, data: &[u8]) {
         self.reply
             .send_ll(&ll::Response::new_ioctl(result, &[IoSlice::new(data)]));
@@ -525,7 +525,7 @@ impl Reply for ReplyPoll {
 }
 
 impl ReplyPoll {
-    /// Reply to a request with the given poll result
+    /// Reply to a request with ready poll events
     pub fn poll(self, revents: u32) {
         self.reply.send_ll(&ll::Response::new_poll(revents));
     }
@@ -649,12 +649,12 @@ impl Reply for ReplyXattr {
 }
 
 impl ReplyXattr {
-    /// Reply to a request with the size of the xattr.
+    /// Reply to a request with the size of an extended attribute
     pub fn size(self, size: u32) {
         self.reply.send_ll(&ll::Response::new_xattr_size(size));
     }
 
-    /// Reply to a request with the data in the xattr.
+    /// Reply to a request with the data of an extended attribute
     pub fn data(self, data: &[u8]) {
         self.reply.send_ll(&ll::Response::new_slice(data));
     }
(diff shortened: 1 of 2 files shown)


src/session.rs
```diff
@@ -198,7 +206,12 @@ pub struct SessionUnmounter {
 
 impl SessionUnmounter {
     /// Unmount the filesystem
+    /// # Errors
+    /// Does not error.
```
cberner, line 210: Let's remove this. I don't think we should document the function as not returning an error, especially given the TODO to make it return an error

src/session.rs
```diff
@@ -247,9 +264,14 @@ impl BackgroundSession {
     /// Create a new background session for the given session by running its
     /// session loop in a background thread. If the returned handle is dropped,
     /// the filesystem is unmounted and the given session ends.
+    /// # Errors
+    /// Does not error.
```
cberner, line 268: Same comment

src/session.rs
```diff
@@ -215,6 +228,10 @@ fn aligned_sub_buf(buf: &mut [u8], alignment: usize) -> &mut [u8] {
 
 impl<FS: 'static + Filesystem + Send> Session<FS> {
     /// Run the session loop in a background thread
+    /// # Errors
+    /// Does not error.
```
cberner, line 232: Same comment

cberner (changes requested):
Had a minor comment about non-errors which I don't think we should document. If it doesn't make sense for a function to return an error, we should change the signature instead

src/session.rs
```diff
@@ -247,21 +259,26 @@ impl BackgroundSession {
     /// Create a new background session for the given session by running its
     /// session loop in a background thread. If the returned handle is dropped,
     /// the filesystem is unmounted and the given session ends.
-    pub fn new<FS: Filesystem + Send + 'static>(se: Session<FS>) -> io::Result<BackgroundSession> {
+    /// # Panics
+    /// Panics if an unmount is already in progress.
+    pub fn new<FS: Filesystem + Send + 'static>(se: Session<FS>) -> BackgroundSession {
         let sender = se.ch.sender();
         // Take the fuse_session, so that we can unmount it
+        // TODO: error instead of panic.
```
cberner, line 267: Hmm, given this TODO it seems like we should leave it as returning a `Result`

src/session.rs
```diff
@@ -262,6 +275,8 @@ impl BackgroundSession {
         })
     }
     /// Unmount the filesystem and join the background thread.
+    /// # Panics
+    /// Panics if the background thread can't be recovered.
```
cberner, line 279: Let's remove this. It's an implementation detail

src/session.rs
```diff
@@ -247,10 +255,15 @@ impl BackgroundSession {
     /// Create a new background session for the given session by running its
     /// session loop in a background thread. If the returned handle is dropped,
     /// the filesystem is unmounted and the given session ends.
+    /// # Errors
+    /// Errors if the mount is poisoned.
```
cberner, line 259: Remove this too

src/session.rs
```diff
@@ -215,6 +221,8 @@ fn aligned_sub_buf(buf: &mut [u8], alignment: usize) -> &mut [u8] {
 
 impl<FS: 'static + Filesystem + Send> Session<FS> {
     /// Run the session loop in a background thread
+    /// # Errors
+    /// Errors if the mount is poisoned.
```
cberner, line 225: Remove this

src/notify.rs
```diff
@@ -85,6 +97,9 @@ impl Notifier {
 
     /// Invalidate the kernel cache for a given directory entry and inform
     /// inotify watchers of a file deletion.
+    /// # Errors
+    /// Returns an error if the notification data is too large.
+    /// Propagates errors due to communicating with the fuse device.
```
cberner, line 102: `Propagates errors due to communicating with the fuse device.`

Let's remove this in all the messages. I don't think it adds useful information. Returning an `io::Result` already explains that it propagates errors

src/reply.rs
```diff
@@ -287,6 +289,10 @@ impl ReplyOpen {
     /// you can pass it as the 3rd parameter of `OpenReply::opened_passthrough()`.  This is done in
     /// two separate steps because it may make sense to reuse backing IDs (to avoid having to
     /// repeatedly reopen the underlying file or potentially keep thousands of fds open).
+    /// # Errors
+    /// Propagates errors due to communicating with the fuse device.
+    /// # Panics
+    /// Panics if this reply object has already been used.
```
cberner, line 295: How can this happen? All the other methods move self

cberner (changes requested):

src/notify.rs
rarensu: I guess that is a fair criticism.

src/reply.rs
rarensu: Currently, it cannot happen. However, it is conceivable that someday, it could become possible. FUSE might force upon us a requirement to support some funny sequence like "reply and then do something else that isn't a reply", that would make it possible to misuse these functions. 

The true, secret reason for this panic is that it isn't possible to define a "consumes self" method on a "box dyn" object. So we have to put it in an option so we can consume the option. Introducing the option introduces the panic. I have a plan in the works that will remove the box dyn from the object stack, which makes this entire train of thought irrelevant.

src/session.rs
cberner: I think we should only document things that are part of the public API, rather than implementation details. That is they're not going to change

src/session.rs
cberner: I usually disable that clippy check in my other projects, because I think it's over eager. We should document errors, but only when they're part of the public API, and in that case "poisoned" needs to be explained clearly so that people understand how the mount would get into that state

src/reply.rs
cberner: Let's remove it then. We shouldn't document things that can't happen, and I generally don't add "Panics" documentation unless it should clearly be part of the public API

src/session.rs
rarensu: I removed most of them, but I didn't remove this one. Joining a background thread can panic, and sometimes should, panic. For example,

- A panic in the session thread would propagate to the main thread on `join()`.
- A deadlocked session thread may or may not cause a panic on `join()`, depending on the OS.

These are both just a normal part of joining a background thread. The underlying call to `thread::join()` even has its own `# Panic` documentation. It would be disingenuous to imply that our function is somehow free from panics.

I did improve the `# Panic` section by clarifying an example.

src/session.rs
rarensu: i decided it makes more sense to change the function implementation, than to try to justify the error condition, so I reverted this change.
See PR #405

src/reply.rs
cberner: Here's an example of how I configured clippy in another crate I maintain: https://github.com/cberner/redb/blob/198df7536cddde80fa7d83b2d80b142d7d89958b/src/lib.rs#L1-L17

cberner (approved):

cberner:
Merged. Thanks for iterating on this. lgtm!

PR #405, opened 2025-09-17
Don't panic

rarensu:
`SessionUnmounter` and `BackgroundSession` have unhandled `.lock().unwrap()` that causes clippy to complain that there should be a `# Panic` section in the function documentation. I think it's more appropriate to handle the poison error, rather than panic.


rarensu:
Additionally, some of those functions had an unused `io::Result` type that can be removed.

rarensu:
To clarify, this PR branches from PR #401 . That one should be merged first.

rarensu:
To clarify, probably, the mutex lock will likely never actually ever become poisoned. There simply isn't any operation currently in use that can fail while holding the lock. So this error handling may never be reached. At least now the warning message in the code clearly demonstrates our intention is to ignore the error.

stepancheg:
> that causes clippy to complain

Clippy does not complain in CI, what's the issue?

Poisoning and panicking seems to be correct behavior. But ignoring and (warning that will be buried in logs) on poison is more dangerous than panic.

PR #409, opened 2025-10-09
Enable cross-compiling with non-Linux host

### LINKED ISSUE #408: build.rs doesn't work with cross-compilation
The current build script (as modified by #398, but the issue existed before that) does not allow some cross-compilation configurations.

As I understand it, the `cfg(target_os = "linux")` kind of checks are meant to conditionally select a behavior or another based on the target platform. But "target" here identifies what the *build script* is built for, not the crate. So, if I'm compiling on Mac I will enter the not-linux branch, even if I'm compiling for Linux.

I'm not sure what the best practice is, but I think some checks in the build script are to be run at runtime, possibly checking against the [`CARGO_CFG_TARGET_OS` env var](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts).

sorcio:
Fixes #408

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -10,33 +10,31 @@ fn main() {
 
     if target_os == "linux" && cfg!(not(feature = "libfuse")) {
         println!("cargo:rustc-cfg=fuser_mount_impl=\"pure-rust\"");
+    } else if target_os == "macos" {
+        pkg_config::Config::new()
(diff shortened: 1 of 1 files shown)


cberner:
That sounds fine

Issue #411, opened 2025-10-16
BackgroundSession::drop() no longer waits for filesystem destroy, causing cleanup issues


anumol-baby:
## Bug Description

After upgrading from fuser **0.14.0** to **0.16.0**, we discovered that `BackgroundSession::drop()` no longer blocks until the filesystem's `destroy()` method is called. This behavioral change causes our worker threads to hang indefinitely because they rely on `destroy()` being called to trigger cleanup.

### Environment
- **fuser version**: 0.16.0 (upgrading from 0.14.0)
- **Rust version**: 1.88.0
- **OS**: Linux (Ubuntu 24.04.2 LTS)
- **Use case**: FUSE filesystem with background worker threads for Azure Blob Storage

### Expected Behavior (v0.14.0)
```rust
{
    let session = BackgroundSession::new(...);
    // ... use session ...
} // <-- drop() blocks here until:
  //     1. Unmount completes
  //     2. destroy() is called on filesystem
  //     3. Background thread finishes
```

Worker threads would detect channel closure (from `destroy()`) and exit cleanly.

### Actual Behavior (v0.16.0)
```rust
{
    let session = BackgroundSession::new(...);
    // ... use session ...
} // <-- drop() returns IMMEDIATELY
  // destroy() is never called (or called much later)
  // Background thread continues running
  // Worker threads hang forever waiting on channels
```

### Root Cause Analysis

The regression stems from this change in `BackgroundSession`:

**v0.14.0**:
```rust
pub struct BackgroundSession {
    _mount: Mount,  // Always present
    guard: Mutex<JoinHandle<Result<()>>>,
}
```

**v0.16.0**:
```rust
pub struct BackgroundSession {
    _mount: Option<Mount>,  // Now optional!
    guard: Mutex<JoinHandle<Result<()>>>,
}
```

**Impact**: When `_mount` is `None` during drop:
- No `Mount::drop()` is triggered
- No blocking occurs waiting for unmount
- `destroy()` is never called on the filesystem
- Background FUSE thread continues running
- Application worker threads hang indefinitely

The `BackgroundSession::join()` method exists:
```rust
pub fn join(self) {
    drop(_mount);
    guard.join().unwrap().unwrap();
}
```

This provides **synchronous cleanup**, but in our architecture:
1. `join()` ensures `destroy()` is called
2. Our `destroy()` implementation is empty (just logging)
3. Worker threads still need explicit signaling to exit

[...]

PR #419, opened 2025-12-10
feat: add optional blocking parameter and unmount mode overrides

khanhtranngoccva:
This attempts to fix issue #411 by adding a method called `set_blocking_umount` on `Session`, `SessionUnmounter` and `BackgroundSession`. 

In all 3 variants of the `Mount` structure, if `blocking_umount` is active, the mount will try to unmount normally, but if it encouters `EBUSY` (`ResourceBusy`) due to open handles, it will wait for 0.5 seconds before retrying. 

The unmounting behavior is also made more uniform - the new behavior is that instead of unpredictable unmount modes happening when `EACCES` is returned, no unmount flags are added. This behavior should be able to be overridden using `set_unmount_options` on session objects.

Additionally, explicitly using unmount() or join() may return an error and allow a busy mount session to be retained if the caller handles it properly, preventing the FS from being abruptly unmounted or the code from blocking.

### DIFF
--- a/src/ll/errno.rs
+++ b/src/ll/errno.rs
@@ -163,80 +163,80 @@ const ALL_RAW_ERRNOS: &[libc::c_int] = &[
     libc::EWOULDBLOCK,
     libc::ENOMSG,
     libc::EIDRM,
-    #[cfg(target_os = "linux")]
+    #[cfg(not(target_os = "macos"))]
     libc::ECHRNG,
-    #[cfg(target_os = "linux")]
+    #[cfg(not(target_os = "macos"))]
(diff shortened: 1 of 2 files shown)


cberner:
Hmm, I see. This is a much more involved change than I'd realized. I'll try to take a look this weekend when I can spend more time on it. Can you get the tests into a passing state? It looks like they're all failing

cberner:
I don't think this PR is moving in a direction where I will be able to review and merge. Please open a series of smaller PRs, as that will be easier to test and review. I'd like the first one to be a test demonstrating the bug (it can be marked `#[ignored]`)

PR #421, opened 2025-12-19
Add FUSE_DEV_IOC_CLONE support for multi-threaded reading

ejc3:
## Summary

Add support for multi-threaded FUSE request processing using FUSE_DEV_IOC_CLONE ioctl.

This enables multiple threads to read FUSE requests in parallel from cloned file descriptors, improving throughput for high-concurrency workloads.

## API

// Primary session handles INIT and runs in main thread
let session = Session::new(fs, mountpoint, &Config::default())?;

// Clone fd for additional reader threads
let cloned_fd = session.clone_fd()?;
let reader_session = Session::from_fd_initialized(fs_clone, cloned_fd, SessionACL::All);
std::thread::spawn(move || reader_session.run());

session.run()?;

## Changes

- Add Session::clone_fd() - clones the FUSE fd using FUSE_DEV_IOC_CLONE ioctl (Linux only)
- Add Session::from_fd_initialized() - creates a session from a cloned fd that skips INIT handshake
- Make Session::run() public for external multi-reader use
- Add conditional handshake skip when proto_version is pre-set

## Tests

Added two integration tests:
- clone_fd_multi_reader - verifies clone_fd() creates a valid fd
- from_fd_initialized_works - tests concurrent readers processing requests

## Platform Support

clone_fd() is only available on Linux (#[cfg(target_os = "linux")]).

### DIFF
--- a/tests/clone_fd_test.rs
+++ b/tests/clone_fd_test.rs
@@ -1,0 +1,191 @@
+//! Test that clone_fd enables true parallel request handling.
+//!
+//! Uses a barrier to prove N threads are executing concurrently.
+//! Without clone_fd, threads serialize on read() and barrier times out.
+
+use std::ffi::OsStr;
+use std::io;
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+use std::sync::Barrier;
+use std::thread;
+use std::time::Duration;
+
+use fuser::Config;
+use fuser::Errno;
+use fuser::FileAttr;
+use fuser::FileType;
+use fuser::Filesystem;
+use fuser::Generation;
+use fuser::INodeNo;
+use fuser::InitFlags;
+use fuser::KernelConfig;
+use fuser::ReplyAttr;
+use fuser::ReplyEntry;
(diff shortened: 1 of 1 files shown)


cberner:
It's this one: https://github.com/cberner/fuser/pull/572

ejc3:
I rebased again. #627 won’t work as is (as mentioned) because it doesn’t support the clone fd ioctl. Otherwise, if that was added, it should work.

ejc3:
Yeah, that makes.
 I was saying that there isn’t a perf improvement if we don’t do the clone though. I will submit pull request on top of 627 that does the clone.

Sent from Yahoo Mail for iPhone


On Sunday, February 1, 2026, 6:06 PM, Stepan Koltsov ***@***.***> wrote:

stepancheg left a comment (cberner/fuser#421)

#627 won’t work as is (as mentioned) because it doesn’t support the clone fd ioctl


multiple event loops can work with and without ioctl. This is how libfuse works — single thread, multiple threads shared fd, or multiple threads clone fd with ioctl.

I started without ioctl, because I presume we should make version without ioctl work first, and then add ioctl on top.

PR #431, opened 2026-01-11
warn(unreachable_pub)

stepancheg:
When reading sources, make more clear what code is public API and what is internals.


cberner:
I love this PR 😀

PR #453, opened 2026-01-14
Add runtime detection for macFUSE 4.x/5.x rename protocol

zah:
## Summary

This patch adds support for macFUSE's extended rename format and the RENAME_SWAP/RENAME_EXCL capabilities, enabling atomic rename operations on macOS.

## Problem

macFUSE supports extended rename operations (`renamex_np` syscall with `RENAME_SWAP` and `RENAME_EXCL` flags) through an extended `fuse_rename_in` structure. However, there's an ABI incompatibility between macFUSE versions:

- **macFUSE 4.x**: Had a bug where it sent the extended 16-byte format unconditionally, regardless of whether RENAME_SWAP/RENAME_EXCL capabilities were negotiated during FUSE_INIT.

- **macFUSE 5.x**: Fixed this behavior - only sends the extended format when capabilities ARE granted; otherwise sends the standard 8-byte format.

This caused fuser to break on macFUSE 5.x when parsing rename requests, as the library wasn't handling both format variations.

### References

- GitHub Issue: [osxfuse/osxfuse#839](https://github.com/osxfuse/osxfuse/issues/839)
- macFUSE kernel header: [fuse_kernel.h](https://github.com/osxfuse/fuse/blob/master/include/fuse_kernel.h)

## Solution

This patch implements a two-pronged approach:

### 1. Capability Negotiation (lib.rs)

Request `FUSE_RENAME_SWAP` and `FUSE_RENAME_EXCL` capabilities during FUSE_INIT:

```rust
#[cfg(target_os = "macos")]
const INIT_FLAGS: u64 = FUSE_ASYNC_READ
    | FUSE_CASE_INSENSITIVE
    | FUSE_VOL_RENAME
    | FUSE_XTIMES
    | FUSE_RENAME_SWAP
    | FUSE_RENAME_EXCL;
```

This tells macFUSE 5.x that we want extended rename support, so it will grant the capabilities and send the extended format.

### 2. Runtime Format Detection (request.rs)

Parse FUSE_RENAME requests with runtime detection of the format:

```
Short format (8 bytes):
  newdir: u64
  <filename bytes follow immediately>

Extended format (16 bytes):
  newdir: u64
  flags: u32
  padding: u32
  <filename bytes follow>
```

The detection heuristic is based on the observation that filenames cannot start with null bytes or ASCII control characters (< 32). By inspecting the first byte after `newdir`:

- If `< 32`: Extended format (flags field contains a small number or zero)
- If `>= 32`: Short format (filename starts with a printable character)

This handles:

1. macFUSE 5.x with capabilities granted (extended format)
2. macFUSE 5.x with capabilities refused (short format)
3. macFUSE 4.x bug behavior (extended format unconditionally)

### `src/lib.rs`

- Added `FUSE_RENAME_SWAP` and `FUSE_RENAME_EXCL` to macOS `INIT_FLAGS`
- Added documentation explaining the macFUSE version differences

### `src/ll/fuse_abi.rs`

- Added `FUSE_RENAME_SWAP` (bit 25) and `FUSE_RENAME_EXCL` (bit 26) capability constants for macOS
- Updated `fuse_rename_in` documentation to explain the variable format

### `src/ll/request.rs`

- Added `flags` field to the `Rename` struct
- Implemented runtime format detection in FUSE_RENAME parsing
- Added comprehensive documentation of the detection algorithm

### `src/ll/argument.rs`

- Added `skip_bytes()` method to `ArgumentIterator` for skipping optional fields
- Added `peek_byte()` method for non-consuming byte inspection

### `src/request.rs`

- Updated dispatch to pass `x.flags()` to `Filesystem::rename()` instead of hardcoded `0`

### DIFF
--- a/src/ll/request.rs
+++ b/src/ll/request.rs
@@ -3,9 +3,9 @@
 //! A request represents information about a filesystem operation the kernel driver wants us to
 //! perform.
 
-use super::fuse_abi::{InvalidOpcodeError, fuse_in_header, fuse_opcode};
+use super::fuse_abi::{fuse_in_header, fuse_opcode, InvalidOpcodeError};
 
-use super::{Errno, Response, fuse_abi as abi};
+use super::{fuse_abi as abi, Errno, Response};
 #[cfg(feature = "serializable")]
 use serde::{Deserialize, Serialize};
 use std::{convert::TryFrom, fmt::Display, path::Path};
@@ -260,11 +260,11 @@ mod op {
     use crate::ll::Response;
 
     use super::{
-        super::{TimeOrNow, argument::ArgumentIterator},
+        super::{argument::ArgumentIterator, TimeOrNow},
         FilenameInDir, Request,
     };
     use super::{
-        FileHandle, INodeNo, Lock, LockOwner, Operation, RequestId, abi::consts::*, abi::*,
+        abi::consts::*, abi::*, FileHandle, INodeNo, Lock, LockOwner, Operation, RequestId,
     };
     use std::{
         convert::TryInto,
@@ -587,6 +587,9 @@ mod op {
         arg: &'a fuse_rename_in,
         name: &'a Path,
         newname: &'a Path,
+        /// Rename flags (macOS only, from extended format).
+        /// On other platforms or when macFUSE sends short format, this is 0.
+        flags: u32,
     }
     impl_request!(Rename<'_>);
     impl<'a> Rename<'a> {
@@ -601,6 +604,14 @@ mod op {
                 dir: INodeNo(self.arg.newdir),
                 name: self.newname,
             }
+        }
+        /// Rename flags (macOS only: RENAME_SWAP, RENAME_EXCL).
+        ///
+        /// On macOS, if FUSE_RENAME_SWAP or FUSE_RENAME_EXCL capabilities were granted
+        /// during init, the kernel sends extended rename requests with flags.
+        /// On other platforms or when macFUSE sends the short format, this returns 0.
+        pub(crate) fn flags(&self) -> u32 {
+            self.flags
         }
     }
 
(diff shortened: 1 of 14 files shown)


src/ll/request.rs
```diff
@@ -1651,12 +1662,74 @@ mod op {
...
+                //   ARE granted; otherwise it sends the short 8-byte format.
+                //
+                // Our solution:
+                // We request FUSE_RENAME_SWAP/FUSE_RENAME_EXCL at INIT (see lib.rs), which
+                // should cause macFUSE 5.x to grant them and send extended format. However,
+                // we use runtime detection as a safety measure because:
+                //   1. The kernel may refuse to grant the capabilities
+                //   2. We need backward compatibility with macFUSE 4.x behavior
+                //
+                // Detection heuristic:
```
cacay, line 1692: Filenames can definitely contain ASCII control sequences. Finder won't allow it, but the terminal and system calls will. The only actual restriction is that filenames cannot contain `NUL` or `/`.

src/ll/request.rs
```diff
@@ -1651,12 +1662,74 @@ mod op {
...
+                //
+                // Background:
+                // macFUSE supports extended rename operations (RENAME_SWAP, RENAME_EXCL)
+                // via flags in the rename request. When these capabilities are negotiated
+                // during FUSE_INIT, macFUSE sends an extended 16-byte fuse_rename_in:
+                //   - newdir: u64 (8 bytes)
+                //   - flags: u32 (4 bytes)
+                //   - padding: u32 (4 bytes)
+                //
+                // The problem:
```
cacay, line 1679: Do we need this detection mechanism? macFUSE 4.x always uses extended. macFUSE 5.x will always use extended since we request the capabilities. I guess there is a theoretical chance the kernel refuses the capabilities, but I don't know if that's possible in practice.

cacay:
@zah I'm looking to support the `renamex_np` callback on macOS. From what I can tell, `fuser` doesn't pipe these flags to `Filesystem::rename`. Any change you'll revisit this PR?

There is now a Linux-only [`RenameFlags`](https://github.com/cberner/fuser/blob/dc83277486fa699c5d8e7f7112a6c0fa2c6a9114/src/rename_flags.rs#L8) struct. Maybe that could be extended to macOS given this PR?

cberner:

PR #456, opened 2026-01-15
Remove feature abi-7-19

stepancheg:
It is supposed to work on mac (I didn't check), and on linux code is supposed to work with every `abi-` feature enabled.


cberner:
I checked that it works on mac

PR #458, opened 2026-01-15
Use ubuntu 24.04 action runner

stepancheg:
Debug new bugs, not old bugs.


cberner:
I think for this to work we need to update to a newer version of this: https://github.com/rfjakob/fuse-xfstests

I haven't looked recently, but I saw some efforts to upstream it

PR #464, opened 2026-01-16
Add fields to fuse_init_in unconditionally

stepancheg:
Both libfuse and macfuse define these fields unconditionally.

libfuse documentation also does not mention that these fields may depend on version
https://github.com/libfuse/libfuse/blob/fa03307c656326733182ab7e30a3cd9e05da43f4/doc/libfuse-operations.txt#L195

libfuse reads the `flags2` field unconditionally
https://github.com/libfuse/libfuse/blob/fa03307c656326733182ab7e30a3cd9e05da43f4/lib/fuse_lowlevel.c#L2711-L2712

So I believe it is OK to not truncate the struct.

Except FreeBSD, that does not declare these fields. https://github.com/freebsd/freebsd-src/blob/1e8c28712aafc9a3339e8a832767a70c08168bdc/sys/fs/fuse/fuse_kernel.h#L759

### DIFF
--- a/src/ll/fuse_abi.rs
+++ b/src/ll/fuse_abi.rs
@@ -16,6 +16,8 @@
 //! - supports ABI 7.19 since FUSE 2.9.1
 //! - supports ABI 7.26 since FUSE 3.0.0
 //!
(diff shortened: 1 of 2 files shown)


stepancheg:
Actually I think we should do what libfuse does (if I understand code correctly, it is not explicit) — zerofill `fuse_init_in`, and then memcpy whatever we read over it.

PR #478, opened 2026-01-16
std -> fuser FileType

stepancheg:
Convenient shortcut


src/lib.rs
```diff
@@ -114,6 +115,30 @@ pub enum FileType {
...
+            FileType::NamedPipe
+        } else if file_type.is_socket() {
+            FileType::Socket
+        } else if file_type.is_char_device() {
+            FileType::CharDevice
+        } else if file_type.is_block_device() {
+            FileType::BlockDevice
+        } else {
+            // fallback
+            FileType::RegularFile
```
cberner: Hmm, I'm nervous about this fallback. It seems like we should either panic or return None

cberner (approved):

PR #479, opened 2026-01-16
Explain blocks, blksize

stepancheg:


src/lib.rs
```diff
@@ -122,7 +122,8 @@ pub struct FileAttr {
     pub ino: u64,
     /// Size in bytes
     pub size: u64,
-    /// Size in blocks
+    /// Allocated size in 512-byte blocks. May be smaller than the actual file size
```
cberner: Is it always 512-byte blocks and not blksize?

stepancheg: All AIs say so, for example https://chatgpt.com/share/696a6f95-3fc8-8002-a5bf-eb6d16dcea6d

cberner: Oh ya, the Linux docs confirm: https://man7.org/linux/man-pages/man3/stat.3type.html

PR #497, opened 2026-01-17
Make offset parameter unsigned

stepancheg:


stepancheg:
I am 95% sure for reasons:
- logically the parameter is unsigned (in both FUSE and pread)
- there's no unsigned version of `off_t` in libc — there is no type that describes file offset and is unsigned
  - similarly `stat.st_size` is also `off_t`, while it can only be non-negative
- `fuse_read_in.offset` type is `uint64_t` in libfuse https://github.com/libfuse/libfuse/blob/fa03307c656326733182ab7e30a3cd9e05da43f4/include/fuse_kernel.h#L826-L828

PR #505, opened 2026-01-18
Add fields to fuse_init_in unconditionally (attempt 2)

stepancheg:
Two commits
- revert #464
- new approach: allocate full `fuse_init_in`, zerofill, and then copy whatever data we have over


cberner:
Thanks! This fixed it

PR #506, opened 2026-01-18
macOS cargo check CI

stepancheg:
For now check it compiles.


cberner:
Jinx :) I just opened #507 too. Not sure my approach of using FSKit is actually going to work though

PR #511, opened 2026-01-18
Allocate buffer as Vec<u64> and bytemuck to cast to bytes

stepancheg:
Simpler logic.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ build = "build.rs"
 
 [dependencies]
 bitflags = "2.9"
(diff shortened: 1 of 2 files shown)


stepancheg:
I'm not happy with current implementation, because it makes buffer size non-determistic: malloc changes, and you got different buffer size.

Probably not a big deal.

cberner:
Ah. I'm happy to merge a fix for just that. I think it's as simple as allocating `BUFFER_SIZE + align<header>`, and then truncating to BUFFER_SIZE?

PR #512, opened 2026-01-18
Extract handshake

stepancheg:
To support `FUSE_DEV_IOC_CLONE`, we need to have this operation
- handshake
- clone the descriptor
- process descriptors from several threads

This change extracts handshake code as preparation to implement this feature.

I'm not 100% sure, but I think `FUSE_OVER_IO_URING` also requires this refactoring: handshake happens over `/dev/fuse` and then rings are created to process messages.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -5,6 +5,7 @@
 //! filesystem is mounted, the session loop receives, dispatches and replies to kernel requests
 //! for filesystem operations under its mount point.
 
+use std::borrow::Cow;
 use std::fs::File;
 use std::io;
 use std::os::fd::AsFd;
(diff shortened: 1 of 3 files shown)


src/session.rs
```diff
@@ -184,6 +196,143 @@ impl<FS: Filesystem> Session<FS> {
...
+            // Parse the request
+            let request = match ll::AnyRequest::try_from(&buf[..size]) {
+                Ok(request) => request,
+                Err(err) => {
+                    error!("{err}");
+                    return Err(io::Error::new(io::ErrorKind::InvalidData, err.to_string()));
+                }
+            };
+
+            debug!("{}", request);
```
cberner: Intentional? Lets add a little more detail, if so

stepancheg: Accidental. Removed now, thanks!

cberner (approved):

PR #534, opened 2026-01-20
Require Send+Sync+'static for Filesystem

stepancheg:
Preparation for multithreaded filesystem: to pass `Arc<impl Filesystem>` into `thread::spawn`, these bounds are needed.


stepancheg:
My thinking is this:
- if you need high performance, you likely need multithreaded processing
- if you don't need high performance, just `Arc<Mutex>` your data

We can support `mount` and `mount_multithreaded` like you described (that would be similar, for example, to `tokio`'s `spawn` and `spawn_local`), but that will be significantly more complex code: for example, I planned to unconditionally spawn threads: one or many depending on configuration, and make single-threaded and multi-threaded implementation the same. Without `Send`, it will be generic code, separate branches to start processing after init, etc etc.

PR #537, opened 2026-01-20
Filesystem::release flags parameter typed

stepancheg:
LLM:

We know that fuse_release_in.flags are the file open flags based on several pieces of evidence in the libfuse source code and documentation:

1. Code Implementation in lib/fuse_lowlevel.c

In the low-level FUSE implementation, the `_do_release` function (lines 1875-1884) explicitly maps the flags from the incoming kernel message (struct fuse_release_in) to the flags field of `struct fuse_file_info`:

```
static void _do_release(fuse_req_t req, const fuse_ino_t nodeid,
			const void *op_in, const void *in_payload)
{
	const struct fuse_release_in *arg = op_in;
	struct fuse_file_info fi;

	memset(&fi, 0, sizeof(fi));
	fi.flags = arg->flags;  // <--- arg->flags is assigned to fi.flags
	fi.fh = arg->fh;
    // ...
```

2. Definition of struct `fuse_file_info` in `include/fuse_common.h`

The definition of struct fuse_file_info (the structure passed to filesystem implementations) explicitly identifies this field as "Open flags":

```
struct fuse_file_info {
	/** Open flags.  Available in open(), release() and create() */
	int32_t flags;
    // ...
```

### DIFF
--- a/examples/passthrough.rs
+++ b/examples/passthrough.rs
@@ -207,7 +207,7 @@ impl Filesystem for PassthroughFs {
         _req: &Request,
         _ino: INodeNo,
         _fh: FileHandle,
-        _flags: i32,
+        _flags: OpenFlags,
         _lock_owner: Option<LockOwner>,
         _flush: bool,
         reply: ReplyEmpty,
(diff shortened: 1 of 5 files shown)


cberner:
Are the flags to `releasedir()` the same? Let's change that too, if so

PR #548, opened 2026-01-20
AsyncFilesystem should be spawn, not block_on

stepancheg:
Calling `block_on` makes `async` useless, preventing concurrent handling of requests.

Previous draft PR was incorrect, this is correct: mt runtime needs to be used, because we are not calling `block_on`.

### DIFF
--- a/src/experimental.rs
+++ b/src/experimental.rs
@@ -1,6 +1,7 @@
 #![allow(missing_docs, missing_debug_implementations)]
 
 use std::ffi::OsStr;
+use std::sync::Arc;
 use std::time::Duration;
 
 use crate::Errno;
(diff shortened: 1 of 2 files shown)


src/experimental.rs
```diff
@@ -123,45 +124,50 @@ impl GetAttrResponse {
...
 }
 
 impl<T: AsyncFilesystem> TokioAdapter<T> {
     pub fn new(inner: T) -> Self {
         Self {
-            inner,
-            runtime: tokio::runtime::Builder::new_current_thread()
+            inner: Arc::new(inner),
+            // Need to use multi-thread runtime because nobody is calling `block_on`
+            // when using `new_current_thread`.
```
cberner: What does this comment refer to?

PR #589, opened 2026-01-28
Fix double umount in umount then drop

stepancheg:


cberner (approved):
Lgtm. But see comment

PR #593, opened 2026-01-28
Read flag parity

khanhtranngoccva:
This PR fixes the `request::op::Read`'s confusing `read_flags` and `flags` properties - the `flags` property should now refer to the file flags, while `read_flags` should now refer to read flags. The code mixed up these two properties (parsing read flags from file flags).


stepancheg (commented):
Fixing read flags is right, but I think we should not be exposing `ReadFlags` to user API, because there's only one flag "lock owner", and we already pass this information as `lock_owner` param.

PR #603, opened 2026-01-30
Do not run apt update in tests

stepancheg:
It already ran in dockerfile, no need to spam repositories on every run.


cberner:
I'm pretty sure these are required. The update in docker may be out of date since that layer of the image can be arbitrarily old

cberner:
I'm fine with making it conditional: check for an Err from `apt install` and then `apt update` and retry. I've debugged this exact issue one too many times in Docker images though :)

PR #605, opened 2026-01-30
Rust instead of sed for fuse.conf in tests

stepancheg:


fuser-tests/src/fuse_conf.rs
```diff
@@ -1,11 +1,37 @@
...
+    Ok(())
+}
 
 pub(crate) async fn fuse_conf_write_user_allow_other() -> anyhow::Result<()> {
-    command_success(["sh", "-c", "echo 'user_allow_other' >> /etc/fuse.conf"]).await
+    let content = fs::read_to_string(FUSE_CONF_PATH)
+        .await
+        .context(format!("Failed to read {FUSE_CONF_PATH}"))?;
+    let lines = content.lines().chain([USER_ALLOW_OTHER]);
+    write_lines(FUSE_CONF_PATH, lines).await
```
wmanley: No need to read the whole thing, tokio supports opening a file in append mode (equivalent to `>>`).

```rust
OpenOptions::new()
        .append(true)
        .open(FUSE_CONF_PATH)
        .await?
        .write(USER_ALLOW_OTHER)
        .await?
```

PR #607, opened 2026-01-30
Check mounts in rust in fuser-tests

stepancheg:


stepancheg:
Changed to parse `mount` output.

PR #620, opened 2026-02-01
Remove Filesystem::destroy

stepancheg:
We have RAII, so it is not very helpful for cleanup, unlike C.

In theory, `destroy` is called before fuse destroy reply is sent, but I cannot find a situation where it could be necessary.

This specific behavior of `destroy` somewhat complicates implementation of multithreading, so I propose just ditch it for time being, and reintroduce later if needed.

Alternative to this PR is #621 — this is complications I'm talking about.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -408,10 +408,6 @@ pub trait Filesystem: Send + Sync + 'static {
         Ok(())
     }
 
-    /// Clean up filesystem.
-    /// Called on filesystem exit.
-    fn destroy(&mut self) {}
-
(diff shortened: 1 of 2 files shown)


cberner:
How will the `Filesystem` be notified of `unmount`? I'm not sure if I've tested, but I assume it receives `destroy()` when unmount is called

cberner:
Ya, so I think we should keep the notification of `umount`. So let's keep `destroy()`, but if you need to defer it being called until some threads are joined or whatever that seems fine

PR #629, opened 2026-02-01
Move examples into own crate

stepancheg:
There is common code in examples (for example, parsing of command line flags or setting up logger). I'ld like to extract this code. The options are:
- move common code into subdirectories in `examples`, so they are not treated as example. This has an issue that everything needs to be public, and dead code is not easy to see, also these are compiled multiple times
- move common code into another crate, like `fuser-example-utils`. This create cyclic dependencies between crates, which is legal is Cargo, but does not look right
- move examples into a separate crate, as proposed by this PR
- do nothing


cberner:
I want to keep the examples in `fuser` because I think it makes them easier to find.

Is it a lot of code that can be reused across examples? My inclination is to do nothing, but if it's a lot of code then option 1 or 2 sounds ok

PR #636, opened 2026-02-02
fuser PR: Add FUSE_DEV_IOC_CLONE for parallel request processing [to merge after #627]

ejc3:
## Summary

  Add `FUSE_DEV_IOC_CLONE` support for true parallel request processing when `n_threads > 1`.

  **Key changes:**
  - Add `Config.clone_fd` option (default `false`) to opt-in to fd cloning
  - When enabled, each worker thread gets its own cloned fd via `FUSE_DEV_IOC_CLONE`
  - Kernel distributes requests across all cloned fds, enabling true parallelism
  - Uses existing `fuse_dev_ioc_clone` from `ll/ioctl.rs`

  ## Background

  Without `clone_fd`: all threads share one fd and serialize on reads - only one thread can have an outstanding kernel request at a time.

  With `clone_fd`: each thread has its own fd, and the kernel distributes requests across them, enabling true parallel request processing.

  ## Changes

  **Core:**
  - `src/channel.rs`: Add `clone_fd()` method using `FUSE_DEV_IOC_CLONE` ioctl
  - `src/session.rs`: Use `clone_fd()` when `config.clone_fd` is true
  - `src/mnt/mount_options.rs`: Add `clone_fd: bool` to `Config`

  **Examples:** Updated to support `--n-threads` and `--clone-fd` flags:
  - `examples/hello.rs` - includes per-thread stats tracking
  - `examples/passthrough.rs`, `examples/null.rs`, etc.

  **Tests:** Updated `fuser-tests` for multi-threaded testing

  ## Usage

  ```rust
  let config = Config {
      n_threads: Some(4),
      clone_fd: true,  // opt-in to FUSE_DEV_IOC_CLONE
      ..Default::default()
  };

  Compatibility

  - Linux 4.5+: Full support with FUSE_DEV_IOC_CLONE
  - Older Linux: Will error if clone_fd: true is set (no silent fallback)
  - macOS/FreeBSD: clone_fd not supported; use n_threads with shared fd
  - API: No breaking changes

  Test plan

  - cargo test passes
  - Manual testing with examples/hello.rs --n-threads 4 --clone-fd
  - Verified multiple fuser-N threads in ps output
  - Debug logs show different threads processing requests

### DIFF
--- a/tests/clone_fd_test.rs
+++ b/tests/clone_fd_test.rs
@@ -5,10 +5,10 @@
 
 use std::ffi::OsStr;
 use std::io;
-use std::sync::atomic::AtomicBool;
-use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::sync::Barrier;
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering;
 use std::thread;
 use std::time::Duration;
 
@@ -81,7 +81,10 @@ impl Filesystem for BarrierFS {
     }
 
     fn lookup(&self, _req: &Request, parent: INodeNo, name: &OsStr, reply: ReplyEntry) {
-        let thread_name = std::thread::current().name().unwrap_or("unknown").to_string();
+        let thread_name = std::thread::current()
+            .name()
+            .unwrap_or("unknown")
+            .to_string();
         eprintln!("Server thread {} got lookup for {:?}", thread_name, name);
 
         // Accept any file starting with "barrier" (barrier0, barrier1, etc.)


src/session.rs
```diff
@@ -245,8 +245,13 @@ impl<FS: Filesystem> Session<FS> {
...
 
         let mut channels = Vec::with_capacity(n_threads);
+        #[cfg(target_os = "linux")]
         for _ in 0..n_threads_minus_one {
-            // TODO: fuse_dev_ioc_clone
+            channels.push(ch.clone_fd()?);
+        }
+        #[cfg(not(target_os = "linux"))]
+        for _ in 0..n_threads_minus_one {
+            // On non-Linux, fall back to sharing the fd (no true parallelism)
```
stepancheg: It is true parallelism, just not as efficient.

tests/clone_fd_test.rs
```diff
@@ -0,0 +1,150 @@
...
+use std::collections::HashSet;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
+use std::thread;
+use std::time::Duration;
+
+use fuser::{Config, Errno, FileType, Filesystem, Generation, INodeNo, ReplyAttr, ReplyEntry, Request, Session};
+
+/// Filesystem that records which thread handled each request
+struct ThreadTrackingFS {
```
stepancheg: This is already done in `examples/hello.rs`. And tests for it live in `fuser-tests` now.

stepancheg: Comment about example.rs and fuser-tests is marked resolved, but there's no reply and no changes?

(Unresolve button does not work.)

ejc3: Sorry about that. added a barrier based test to prove the kernel is spreading requests across fd's in parallel.

src/mnt/mount_options.rs
```diff
@@ -15,6 +15,10 @@ pub struct Config {
     pub acl: SessionACL,
     /// Number of event loop threads. If unspecified, one thread is used.
     pub n_threads: Option<usize>,
+    /// Use FUSE_DEV_IOC_CLONE to give each worker thread its own fd.
+    /// This enables true parallel request processing from the kernel.
+    /// Requires Linux 4.5+. Falls back to shared fd on older kernels.
+    pub clone_fd: bool,
```
cberner: What's the rationale for this being a config option? Are there cases where users won't want to use clone_fd?

PR #646, opened 2026-02-07
FUSE_DEV_IOC_CLONE

stepancheg:
This is re-submit of #636 with a few tweaks:

- tests converted to `fuser-tests`
- error on attempt to use `clone_fd` on non-Linux
- remove excessive comments
- fix incorrect comment about parallel processing
- small code cleanups

Closes: #636

### DIFF
--- a/examples/common/args.rs
+++ b/examples/common/args.rs
@@ -19,6 +19,12 @@ pub struct CommonArgs {
     /// Number of threads to use
     #[clap(long, default_value_t = 1)]
     pub n_threads: usize,
+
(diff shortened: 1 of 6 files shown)


stepancheg:
> how threads blocked on waiting for FD can do work without clone

Thread A: reading message from kernel
Thread B: blocked reading
Thread C: blocked reading

Thread A: parsing request
Thread B: reading next message from kernel
Thread C: blocked reading

Thread A: invoking user handler
Thread B: parsing request
Thread C: reading next message from kernel

PR #648, opened 2026-02-12
[draft] Add fallible unmounting

khanhtranngoccva:
This serves as a more manageable version of #419. I cherry-picked the test cases and the CI modifier that allows these tests to execute. The tests cannot compile as of now, since the PR changed the API interface of `umount_and_join`.


khanhtranngoccva:
@cberner Please let me know what behavior you want because currently, the unmounting behavior is non-deterministic and not well-defined in Linux. 

As far as I know, the current libfuse3 implementations is also leaky and not well defined as mentioned by the previous PR(s), as it does not call fuse_session_destroy if on root.

PR #659, opened 2026-02-25
Make `mount_point` argument positional in `examples/simple.rs`

numbleroot:
Simple change to fix the `TODO` stated for CLI argument `mount_point` in `examples/simple.rs`: To make the argument positional, as is the case for the other examples. This is achieved simply by removing the `#[clap(long, default_value = "")]` line right above `mount_point`, as fields without attributes define positional arguments by default. And by adjusting the tests that use the `simple` example.

Thanks for building and maintaining this useful crate!


cberner:

PR #660, opened 2026-02-25
Allow for raw marshalling of BackingIds

Popax21:
Adds methods to convert `BackingIds` from/to their raw `backing_id` values, which allows for marshalling of backing file references across process boundaries.

Currently, the Linux kernel still requires `CAP_SYS_ADMIN` to use FUSE passthrough.
Since this is a highly privileged capability, FUSE filesystem drivers may wish to move this part of their FS into a privileged host process, while continuing to run the rest of the FS at a lower privilege level.
However, this requires the ability to move a `BackingId` reference across process boundaries; this PR adds two methods, `BackingId::into_raw()` and `ReplyOpen::wrap_backing()`, to convert the safe Rust wrapper to/from the underlying raw `backing_id` value respectively, which the user may then e.g. send across process boundaries.

`ReplyOpen::wrap_backing()` is marked as unsafe, even though as far as I can tell misuse of this method does not result in anything else than subsequent operations on the `BackingId` erroring out.
However just to be safe / clearly communicate the "raw-ness" of the operation I have still decided to mark it as unsafe anyway.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,3 +85,6 @@ name = "ioctl"
 
 [[example]]
 name = "passthrough"
+
+[[example]]
+name = "passthrough_fork"
--- a/Makefile
+++ b/Makefile
@@ -63,8 +63,9 @@ mount_tests:
 	 fuser:mount_tests_libfuse3 bash -c "cd /code/fuser && cargo run -p fuser-tests -- linux-mount-libfuse3"
 
 test_passthrough:
-	cargo build --example passthrough
+	cargo build --example passthrough --example passthrough_fork
 	sudo tests/test_passthrough.sh target/debug/examples/passthrough
+	sudo tests/test_passthrough.sh target/debug/examples/passthrough_fork
 
 test: pre mount_tests pjdfs_tests xfstests
 	cargo test
--- a/examples/passthrough_fork.rs
+++ b/examples/passthrough_fork.rs
@@ -1,0 +1,335 @@
+// This example requires fuse 7.40 or later. Run with:
+//
+//   cargo run --example passthrough_fork /tmp/foobar
+
+mod common;
+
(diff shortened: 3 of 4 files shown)


Popax21:
Fixed the rustc lint build failure (seems like it only appears on nightly, and I ran my pre-commit checks on stable; oops...)

cberner:
I wondered if we'd run into issues with nightly rustfmt regressions. I'll see if I can revert it to stable. I believe it's only needed for one format option, that's nice to have but not really required

Popax21:
> I merged #661
> 
> Can you try rebasing on master?

Done; also removed the extra parenthesis again since hopefully that lint should no longer trigger a false positive.

src/lib.rs
```diff
@@ -1054,8 +1053,7 @@ pub fn mount<FS: Filesystem, P: AsRef<Path>>(
     mountpoint: P,
     options: &Config,
 ) -> io::Result<()> {
-    check_option_conflicts(options)?;
```
cberner: Please submit this in a separate PR, since it isn't related to the BackingId feature, as far as I can tell

cberner (changes requested):
This feature still needs a test that runs as part of CI. It should test that it works as expected: backing ids can be passed from a privileged process to an unprivileged process and used there

Popax21:
> This feature still needs a test that runs as part of CI. It should test that it works as expected: backing ids can be passed from a privileged process to an unprivileged process and used there

Done; the `passthrough_fork.rs` example now runs as part of CI, similarly to how `passthrough.rs` already does (I was initially under the assumption that the CI runner does not have access to `CAP_SYS_ADMIN`, so I did not implement this at first; sorry for the mistake). Also addressed the other PR reviews.

Issue #164, opened 2021-06-22
New API using async


cberner:
It would be nice to have an easier to use `Filesystem` API that also uses Rust's `async` feature

paradigm:
[Tokio-uring 0.1.0 was just announced](https://tokio.rs/blog/2021-07-tokio-uring) and, where available, is likely to show significant performance improvements for specifically filesystem I/O that would be common with FUSE.  It may be worth consideration for this issue.  The main catch is it requires a newer Linux kernel; otherwise, a different library with a different interface (e.g. normal Tokio) is needed.

leakingtapan:
I'm working with an API that is generated from protobuf using tonic. The generated APIs are `async` functions. I would be great if I can call into the APIs from fuser's `Filesystem` API without translating the async calls to sync calls. Is there any plan on solving this issue any time soon?

@paradigm any pointers on how to integrate with the new Tokio-uring? is that something coming out of the box that can be enable from tokio at high level with some configurations, or is that some low level API needs to be written before hand, something like read `/dev/fuse` file with io_uring?

paradigm:
> @paradigm any pointers on how to integrate with the new Tokio-uring?

I haven't actually programmed anything in it, just read about it at a high level; I don't have enough experience with it to provide much guidance.

messense:
There is https://github.com/3442/blown-fuse , but it's linux only.

iwanbk:
> Check out https://github.com/Sherlock-Holo/fuse3 . Seems linux only too.

@Mivik fuse3 v0.8.1 (two days after your comment) added macos support, i haven't tried it yet tho

cberner:
Yep. I just haven't gotten around to working on its performance. Feel free to submit PRs if you'd like to work on the async API

PR #663, opened 2026-03-14
Cleanup async interface and map 1:1 with Filesystem trait

mhambre:
This PR takes the experimental async interface and cleans it up such that expansion can start to easily add new methods by mapping 1:1 with the `Filesystem` trait. This adds a `call_fut!` macro that sends the task to the tokio runtime, and automatically clones the request and a reference to the runtime to reduce boilerplate.

This is made with the goal in mind of next steps being expansion to include more mutation oriented operations, and orienting the async capabilities in a way that even further down the line this can be backed with [AsyncFD](https://docs.rs/tokio/latest/tokio/io/unix/struct.AsyncFd.html), setting up an Async-oriented EventLoop, etc for full end-to-end async.

If this is accepted I can also take the initiative of starting to add the other missing FUSE methods to the async trait and building up a simple_async.rs example FS + tests.


mhambre:
#164

mhambre:
@cberner Any thoughts on this when you get some time would be appreciated :)

cberner:
I don't like this interface for async. One of my goals with the experimental async interface is to remove the need for Reply objects

cberner:
I don't like this interface. I intentionally removed the Reply objects in the experimental async interface

mhambre:
> I don't like this interface. I intentionally removed the Reply objects in the experimental async interface

I appreciate your response and clarification. What would your thoughts be then on biting the bullet and decoupling it from the current Filesystem trait pretty much entirely (since as it stands, regardless of where it is Reply, will be blocking) and moving to the Result based approach for [AsyncFD](https://docs.rs/tokio/latest/tokio/io/unix/struct.AsyncFd.html). I can also keep it the more minimal interface you were going for in the initial implementation. I know this is a bit of a step away from libfuse and a significant addition, but if we add async, I agree with the sentiment of being more rustic with Results since we aren't dealing with the same limitations. I'd obviously keep it as tightly scoped as possible for the initial PoC.

cberner:
Ya, that sounds like it's aligned with the direction I want to take it. However, that's going to be a very large change, and I don't have time to review that right now. I'd encourage you to make a fork instead, and then circle back once you have it working and benchmark the performance

PR #667, opened 2026-04-09
Add `ReplyEntry::entry_with_ttls` for distinct TTLs

AM5800:
Currently, `ReplyEntry::entry` forces the same TTL for both the directory entry and its attributes. 

In our VFS in certain situations we need to audit all access,  including stat, so we set attribute cache duration to zero. 

Setting entry cache to zero at the same time completely destroys  performance, so I would like to add an ability to configure these  timeouts separately.

### DIFF
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -243,6 +243,33 @@ impl ReplyEntry {
         ));
     }
 
+    /// Reply to a request with the given entry.
+    ///
+    /// Use this function instead of `ReplyEntry::entry` if the
+    /// entry TTL needs to be different from the attribute TTL.
+    ///
+    /// * `attr_ttl` - Time-to-live for the file attributes. Controls how long the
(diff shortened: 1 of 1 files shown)


src/reply.rs
```diff
@@ -243,6 +243,26 @@ impl ReplyEntry {
         ));
     }
 
+    /// Reply to a request with the given entry.
+    ///
+    /// Use this function instead of `ReplyEntry::entry` if the
+    /// entry TTL needs to be different from the attribute TTL.
```
cberner, line 249: Can you add an explanation of the two different TTLs? Even I have no idea what they really do, lol

cberner (changes requested):
One comment. Otherwise lgtm

src/reply.rs
AM5800: Expanded comments, thanks!

cberner (approved):

PR #678, opened 2026-05-04
Export ForgetOne to allow implementing Filesystem::batch_forget

olvrlrnz:
Implementors of `fuser::Filesystem` cannot implement `batch_forget` because `ForgetOne` is private. This commit makes it public.


cberner:

PR #679, opened 2026-05-04
feat(session): add from_fds for externally pre-cloned fds

XciD:
## Summary

Adds `Session::from_fds(filesystem, primary_fd, extra_fds, acl, config)` for callers that receive `/dev/fuse` fds cloned by a privileged helper (e.g. a CSI driver sending fds over `SCM_RIGHTS`).

When `extra_fds` is non-empty:
- Worker count = `1 + extra_fds.len()` (`config.n_threads` is ignored).
- Workers use the supplied fds directly; `config.clone_fd` and the internal `clone_fd()` path (which opens `/dev/fuse` and so requires `CAP_SYS_ADMIN`) are skipped.

Each extra fd must already be bound to the same FUSE connection as the primary (`FUSE_DEV_IOC_CLONE` ioctl); `FUSE_INIT` still happens on the primary during `handshake()`.

`from_fd` is preserved as a thin wrapper over `from_fds` with no extras.

## Motivation

In some deployments (e.g. Kubernetes CSI drivers), an unprivileged process should not hold `CAP_SYS_ADMIN` and so cannot itself open `/dev/fuse` to clone the session fd. A privileged sidecar opens `/dev/fuse`, performs `FUSE_DEV_IOC_CLONE`, and passes the resulting fds to the worker over a Unix socket. This API lets the worker drive a multi-threaded session over those externally-cloned fds without ever opening `/dev/fuse` itself.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -128,6 +128,12 @@ pub struct Session<FS: Filesystem> {
     pub(crate) filesystem: FilesystemHolder<FS>,
     /// Communication channel to the kernel driver
     pub(crate) ch: Channel,
+    /// Extra channels over pre-cloned /dev/fuse fds. When non-empty, `run()`
+    /// spawns one worker per channel (plus the primary) instead of calling
+    /// `clone_fd()` on Linux. Set by `Session::from_fds` for callers that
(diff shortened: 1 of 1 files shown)


cberner:

PR #689, opened 2026-06-19
Lookup-time passthrough initialization

khanhtranngoccva:
Since a FUSE inode only allows all opens of an inode to be opened as 1 passthrough ID or none at all, this PR aims to allow eager initialization of the passthrough ID for an inode at lookup time, which is much easier to do compared with doing the same thing in open() or create().


cberner:

PR #690, opened 2026-06-19
Allowing 'static capture of Request objects

khanhtranngoccva:
Request objects cannot be cloned, so there is no way to upgrade the Request references to an owned object for capture in structured error handling.


cberner:

PR #691, opened 2026-06-23
Allow user to set thread name

khanhtranngoccva:
This PR allows library users to set their own thread names. This allows certain scenarios like setting thread names to the same value for aggregating in a flamegraph.


cberner:

PR #692, opened 2026-07-01
Add sync_fs() to service FUSE_SYNCFS

edmc-ss:
## Summary
Adds support for servicing `FUSE_SYNCFS` — the opcode the kernel issues for `syncfs(2)`/`sync(2)` — plus the opt-in `FUSE_HAS_SYNCFS` INIT capability that lets a privileged `/dev/fuse` server receive it.

- New `Filesystem::sync_fs(&self, req, ino, ReplyEmpty)` trait method; defaults to `ENOSYS`, which the kernel treats as a permanent success (same convention as `fsync`/`statfs`).
- `FUSE_SYNCFS = 50` opcode, a header-only `SyncFs` request, its `Operation` variant, and request parsing/dispatch/`Display`.
- `InitFlags::FUSE_HAS_SYNCFS = 1 << 43` (protocol 7.46), requested by a server via `KernelConfig::add_capabilities()`.

## Dependency on the Linux VFS fuse driver (please read)
This is the important caveat. Although the `FUSE_SYNCFS` opcode has existed for a while (added for virtiofs), the in-kernel fuse driver only *propagates* `syncfs()`/`sync()` to the userspace server (`fc->sync_fs`) for **virtiofs** and **fuseblk** mounts. A plain `/dev/fuse` server never receives it today — regardless of what it advertises — because an untrusted server could stall `sync()` indefinitely.

Consequences:
- On **current released kernels**, this PR immediately benefits **fuseblk** `fuser` servers (they already receive `FUSE_SYNCFS`), and is a harmless no-op for plain servers.
- For **plain `/dev/fuse` servers** to receive it, the kernel needs the in-flight opt-in: a new `FUSE_HAS_SYNCFS` INIT flag, honored only when the server opened `/dev/fuse` with `CAP_SYS_ADMIN` in the initial user namespace (the same privilege that mounting virtiofs/fuseblk already requires). That series — *"fuse: allow FUSE_SYNCFS for privileged userspace servers"* — is on LKML ([patch](https://lists.openwall.net/linux-kernel/2026/06/19/1124), [LWN summary](https://lwn.net/Articles/1078768/)) and **not yet merged**. The `1 << 43` bit and protocol minor 7.46 here match that patch's UAPI.

Because `KernelConfig::add_capabilities(InitFlags::FUSE_HAS_SYNCFS)` returns `Err` when the running kernel doesn't offer the flag, a server can cleanly detect an unsupporting kernel and fall back (e.g. to an out-of-band flush) rather than silently never being called.

## Design notes
- `FUSE_KERNEL_MINOR_VERSION` is intentionally left at 40. The kernel gates this flag on the opener's privilege rather than the negotiated minor, and it rides in the high-32 `flags2` (already carried via `FUSE_INIT_EXT`), so it works without claiming the intervening 7.41–7.45 features `fuser` doesn't yet implement. Happy to bump the minor if you'd prefer.
- `SyncFs` is modeled on `StatFs`: filesystem-wide and header-only (the wire `fuse_syncfs_in` is a single reserved `u64` padding word, so there's no argument to expose).
- Mirrors libfuse's low-level `syncfs` op, which likewise replies `ENOSYS` when unimplemented.

## Testing
`cargo test` (including a new `FUSE_SYNCFS` parse test), `cargo clippy --all-targets`, `cargo fmt --check`, and `cargo doc` (with `-D rustdoc::broken_intra_doc_links`) all pass on Linux (aarch64, toolchain 1.85). Not exercised against a live mount, since the privileged-plain-server kernel path isn't in a shipping kernel yet (see above).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # FUSE for Rust - Changelog
 
 ## 0.18.0 - Unreleased
+* Add `Filesystem::sync_fs()` to service `FUSE_SYNCFS` (`syncfs(2)`/`sync(2)`), plus the
+  `InitFlags::FUSE_HAS_SYNCFS` opt-in capability a privileged server requests via
+  `KernelConfig::add_capabilities()`
 * Remove deprecated feature flags `abi-*`
 * Rename `mount2()` to `mount()`
 * Rename `spawn_mount2()` to `spawn_mount()`
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -777,6 +777,22 @@ pub trait Filesystem: Send + Sync + 'static {
         reply.statfs(0, 0, 0, 0, 0, 512, 255, 0);
     }
 
+    /// Synchronize the file system.
+    ///
+    /// Called when userspace issues `syncfs(2)` (or `sync(2)`) against the mount,
+    /// giving the server a chance to flush any state it has buffered before
+    /// replying. The kernel only forwards this once the server has opted in by
+    /// requesting [`InitFlags::FUSE_HAS_SYNCFS`] in [`Filesystem::init`]
+    /// (via [`KernelConfig::add_capabilities`]), and only honors that opt-in for
+    /// servers that opened `/dev/fuse` with `CAP_SYS_ADMIN` in the initial user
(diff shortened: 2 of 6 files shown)


cberner:

PR #696, opened 2026-07-17
feat: add low level interrupts

khanhtranngoccva:
This method adds the ability to implement interrupts, and interrupts run in a dedicated thread pool.

This is critical when dealing with cancelling locks that are blocking or stuck. I have just finished a port of ulockmgr that can be used here.


cberner:

PR #703, opened 2026-07-24
Fix inverted POLLERR check in is_mounted()

cberner:
## Problem

`is_mounted()` in `src/mnt/mod.rs` polls `/dev/fuse` with no requested events and treats `POLLERR` as "mounted". The kernel's `fuse_dev_poll` reports `POLLERR` precisely when the connection is *dead* (unmounted or aborted), so the function returned `true` in every reachable case: `Ok(0)` while mounted, and `POLLERR` after unmount. Verified empirically: after an external `umount`, `is_mounted()` still returned `true`.

This defeated the guard in `fuse_pure::MountImpl::umount_impl()`, whose comment explains it exists to avoid "a race with a newly mounted filesystem living at the same mountpoint". After a user unmounted externally (the most common way a session ends), teardown would proceed to `umount(2)` / `fusermount -u -q -z` on the path anyway - and if a new filesystem had been mounted there in the meantime (e.g. a daemon restart where the new instance mounts before the old process exits), fuser could unmount that unrelated, live filesystem. It also produced spurious "Unmount failed" warnings in the ordinary case.

## Fix

Invert the `Ok(1)` arm: the connection is mounted only if `POLLERR` is *not* set. This matches libfuse's `fuse_kern_unmount()`, which skips the unmount when `poll()` reports `POLLERR`.

Behavioral note: as in libfuse, a connection that was aborted via `/sys/fs/fuse/connections/*/abort` (dead connection, mountpoint still mounted) now also skips the teardown unmount.

## Testing

* New regression test `mnt::test::mount_unmount_external`: mounts, unmounts externally (via `fusermount3`/`fusermount`, falling back to `umount(2)` when running as root), and asserts `is_mounted()` becomes false. The test fails against the previous code and passes with the fix.
* The `mount_unmount` name prefix keeps it covered by the existing `--skip=mnt::test::mount_unmount` substring filters used by the FreeBSD CI and `make test_macos`, where unprivileged mounting is unavailable.
* `cargo test --lib` (54 tests), `cargo fmt --check`, and `cargo clippy --all-targets` with `RUSTFLAGS=--deny warnings` (default features and `--no-default-features`) all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TfPeUUeQKYjo26Y9mQG9sA

---
_Generated by [Claude Code](https://claude.ai/code/session_01TfPeUUeQKYjo26Y9mQG9sA)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # FUSE for Rust - Changelog
 
+## Unreleased
+* Fix inverted mounted-check during session teardown: after the filesystem had already been
+  unmounted externally, fuser would attempt to unmount the mountpoint again, which could
+  unmount an unrelated filesystem mounted at the same path in the meantime
+
 ## 0.18.0 - 2026-07-22
 * Remove deprecated feature flags `abi-*`
 * Rename `mount2()` to `mount()`
--- a/src/mnt/mod.rs
+++ b/src/mnt/mod.rs
@@ -207,7 +207,9 @@ fn is_mounted(fuse_device: &DevFuse) -> bool {
         let res = poll(slice::from_mut(&mut poll_fd), PollTimeout::ZERO);
         break match res {
             Ok(0) => true,
-            Ok(1) => poll_fd
+            // The kernel reports POLLERR on /dev/fuse if and only if the connection
+            // is no longer alive, i.e. the filesystem has been unmounted or aborted
+            Ok(1) => !poll_fd
                 .revents()
                 .is_some_and(|r| r.contains(PollFlags::POLLERR)),
             Ok(_) => unreachable!(),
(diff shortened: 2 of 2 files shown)


cberner:
CI note: `test (mount_tests)` failed with `stats_per_thread.len() == n_threads (3 vs 2)` in the multi-threaded stats check. This is a pre-existing flake, not caused by this diff - the same job failed on master at `72fafae` ([run 29891505897](https://github.com/cberner/fuser/actions/runs/29891505897)) in the same phase, there as `Failed to parse stats line: cannot parse integer from empty string`. I've re-run the failed jobs.

Root cause of the flake, for reference: `hello.rs` serves `stats-per-thread` with zero attr TTL while its content grows, so `read_to_string` in fuser-tests can issue a second read(2) after the size was re-fetched, stitching together fragments of two different snapshots - yielding an empty line or an extra line. Padding the counters to a fixed width in `stats_content()` (constant file size) would eliminate the race; happy to send that as a separate PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

